PR #1, opened 2021-09-11
Support initrd loading to boot Linux

RinHizakura:
Does it make sense to support new features for this minimalist type 2 hypervisor? I have tried to support the loading of initrd for `kvm-host`. Until now,  by setting the e820 memory map and load initrd to the right place, this patch makes the booting of Linux on kvm-host more progress. 

The following image shows the last message for the booting of Linux after this patch. What needs to be mentioned is that there's still many warning/error happened during the booting process, we should fix them in the future.
image

By the way,  to support initrd for Linux, I tried to turn on the `CONFIG_BLK_DEV_INITRD` and `CONFIG_BLOCK ` in `tiny.config` and use the config to compile Linux first. But it still can't work when only enable them. Some reports like `Kernel panic - not syncing: No init found. Try passing init= option to kernel.`  still happen. So I turn to use the default config for x86 (by make x86_64_defconfig) in this patch.


scripts/build-initrd.sh
```diff
@@ -0,0 +1,36 @@
...
+    tar -xjf ${OUT}/${BUSYBOX_ARCHIVE} -C ${OUT} || exit 1
+}
+
+function build_busybox()
+{
+   echo "Building Busybox..."
+   cp -f scripts/busybox.config ${OUT}/busybox-${BUSYBOX_VER}/.config
+   pushd ${OUT}/busybox-${BUSYBOX_VER}
+   make -j `nproc` && make install
+   cd _install
```
jserv: Split the following commands into new function. Let's say, `generate_rootfs`.

scripts/busybox.config
```diff
@@ -0,0 +1,1205 @@
+#
```
jserv: Remove the comments in the head. No busybox versioning should be kept.

jserv:
> By the way, to support initrd for Linux, I tried to turn on the `CONFIG_BLK_DEV_INITRD` and `CONFIG_BLOCK ` in `tiny.config` and use the config to compile Linux first. But it still can't work when only enable them. Some reports like `Kernel panic - not syncing: No init found. Try passing init= option to kernel.` still happen. So I turn to use the default config for x86 (by make x86_64_defconfig) in this patch.

You have to enable the following configurations:
* `CONFIG_BLK_DEV_INITRD`
* `CONFIG_BINFMT_ELF`
* `CONFIG_BINFMT_SCRIPT`
* `CONFIG_PROC_FS`
* `CONFIG_SYSFS`

scripts/build-initrd.sh
```diff
@@ -0,0 +1,36 @@
...
+   mv linuxrc init
+   mkdir -p etc/init.d
+   cd etc/init.d
+   echo -e "#!/bin/sh\nmount -t devtmpfs devtmpfs /dev\nmkdir -p /dev/pts\nmount -vt devpts -o gid=4,mode=620 none /dev/pts" > rcS
+   chmod 755 rcS
+   cd ../../../_install
+   find . | cpio -o --format=newc > root_fs.cpio
```
jserv: Change this command to
```shell
find . | cpio -o --format=newc > ../../rootfs.cpio
```

Always use "rootfs" instead of "root_fs" for consistency.

RinHizakura:
> You have to enable the following configurations:
> 
> * `CONFIG_BLK_DEV_INITRD`
> * `CONFIG_BINFMT_ELF`
> * `CONFIG_BINFMT_SCRIPT`
> * `CONFIG_PROC_FS`
> * `CONFIG_SYSFS`

Here's the result until the last message after enabling the configurations. Look like it works as expected.

image

jserv:
> Here's the result until the last message after enabling the configurations. Look like it works as expected.

Great. Then, we need the device emulation for 8280 UART serial to make console work. See [gokvm](https://github.com/bobuhiro11/gokvm) for its neat implementation.

PS: Don't attach the screenshot which contains only plain text or kernel messages.

scripts/busybox.config
```diff
@@ -0,0 +1,1201 @@
...
+
+#
+# Print Utilities
+#
+CONFIG_LPD=y
+CONFIG_LPR=y
+CONFIG_LPQ=y
+
+#
+# Mail Utilities
```
jserv, line 947: Remove mail utilities since we lack of networking.

scripts/busybox.config
```diff
@@ -0,0 +1,1201 @@
...
+CONFIG_ASH_IDLE_TIMEOUT=y
+CONFIG_ASH_MAIL=y
+CONFIG_ASH_ECHO=y
+CONFIG_ASH_PRINTF=y
+CONFIG_ASH_TEST=y
+CONFIG_ASH_HELP=y
+CONFIG_ASH_GETOPTS=y
+CONFIG_ASH_CMDCMD=y
+CONFIG_CTTYHACK=y
+CONFIG_HUSH=y
```
jserv: We only need `ash`. Remove `hush` for space reduction.

scripts/busybox.config
```diff
@@ -0,0 +1,1201 @@
...
+CONFIG_FEATURE_KILL_DELAY=0
+CONFIG_FEATURE_INIT_SCTTY=y
+CONFIG_FEATURE_INIT_SYSLOG=y
+CONFIG_FEATURE_INIT_QUIET=y
+# CONFIG_FEATURE_INIT_COREDUMPS is not set
+CONFIG_INIT_TERMINAL_TYPE="linux"
+CONFIG_FEATURE_INIT_MODIFY_CMDLINE=y
+
+#
+# Login/Password Management Utilities
```
jserv, line 496: You can remove the password management utilities because the environment is intended to be single user.

scripts/busybox.config
```diff
@@ -0,0 +1,1201 @@
...
+CONFIG_FEATURE_LS_FILETYPES=y
+CONFIG_FEATURE_LS_FOLLOWLINKS=y
+CONFIG_FEATURE_LS_RECURSIVE=y
+CONFIG_FEATURE_LS_WIDTH=y
+CONFIG_FEATURE_LS_SORTFILES=y
+CONFIG_FEATURE_LS_TIMESTAMPS=y
+CONFIG_FEATURE_LS_USERNAME=y
+CONFIG_FEATURE_LS_COLOR=y
+CONFIG_FEATURE_LS_COLOR_IS_DEFAULT=y
+CONFIG_MD5SUM=y
```
jserv: Checksum tools are useless at the moment.

scripts/busybox.config
```diff
@@ -0,0 +1,1201 @@
...
+CONFIG_TRUE=y
+CONFIG_TRUNCATE=y
+CONFIG_TTY=y
+CONFIG_UNAME=y
+CONFIG_UNAME_OSNAME="GNU/Linux"
+CONFIG_BB_ARCH=y
+CONFIG_UNIQ=y
+CONFIG_UNLINK=y
+CONFIG_USLEEP=y
+CONFIG_UUDECODE=y
```
jserv: Ditto. Remove `uudecode`, `base32`, `base64`, and `uuencode`.

scripts/busybox.config
```diff
@@ -0,0 +1,1201 @@
...
+CONFIG_BASENAME=y
+CONFIG_CAT=y
+CONFIG_FEATURE_CATN=y
+CONFIG_FEATURE_CATV=y
+CONFIG_CHGRP=y
+CONFIG_CHMOD=y
+CONFIG_CHOWN=y
+CONFIG_FEATURE_CHOWN_LONG_OPTIONS=y
+CONFIG_CHROOT=y
+CONFIG_CKSUM=y
```
jserv: Ditto. Remove `cksum` and `crc32`.

RinHizakura:
> PS: Don't attach the screenshot which contains only plain text or kernel messages.

Sorry for doing this.

The configs are changed accordingly, sincerely ask you to help check.

scripts/build-initrd.sh
jserv: ?

src/kvm-host.c
```diff
@@ -0,0 +1,27 @@
+#if !defined(__x86_64__) || !defined(__linux__)
```
jserv: The `#if` guard should exist in file `src/vm.c` which is the real KVM implementation.

Makefile
```diff
@@ -17,15 +17,15 @@ else
     VECHO = @printf
 endif
 
-OBJS := kvm-host.o
+OBJS := kvm-host.o vm.o
```
jserv: Move `vm.o` before `kvm-host.o` so that `#if` guard can take place before generic `kvm-host`.

scripts/busybox.config
```diff
@@ -0,0 +1,1148 @@
...
+CONFIG_UNXZ=y
+CONFIG_XZCAT=y
+CONFIG_XZ=y
+CONFIG_BZIP2=y
+CONFIG_BZIP2_SMALL=8
+CONFIG_FEATURE_BZIP2_DECOMPRESS=y
+CONFIG_CPIO=y
+CONFIG_FEATURE_CPIO_O=y
+CONFIG_FEATURE_CPIO_P=y
+CONFIG_DPKG=y
```
jserv: `dpkg` is not necessary. We never install packages in such environment.

scripts/busybox.config
```diff
@@ -0,0 +1,1148 @@
...
+#
+# Console Utilities
+#
+CONFIG_CHVT=y
+CONFIG_CLEAR=y
+CONFIG_DEALLOCVT=y
+CONFIG_DUMPKMAP=y
+CONFIG_FGCONSOLE=y
+CONFIG_KBD_MODE=y
+CONFIG_LOADFONT=y
```
jserv: Remove `loadfont` and `setfont` because we don't have real display device.

scripts/busybox.config
```diff
@@ -0,0 +1,1148 @@
...
+CONFIG_SETFONT=y
+CONFIG_FEATURE_SETFONT_TEXTUAL_MAP=y
+CONFIG_DEFAULT_SETFONT_DIR=""
+
+#
+# Common options for loadfont and setfont
+#
+CONFIG_FEATURE_LOADFONT_PSF2=y
```
jserv: Ditto. Remove the entries related to `loadfont`.

scripts/busybox.config
```diff
@@ -0,0 +1,1148 @@
...
+CONFIG_RESIZE=y
+CONFIG_FEATURE_RESIZE_PRINT=y
+CONFIG_SETCONSOLE=y
+CONFIG_FEATURE_SETCONSOLE_LONG_OPTIONS=y
+CONFIG_SETKEYCODES=y
+CONFIG_SETLOGCONS=y
+CONFIG_SHOWKEY=y
+
+#
+# Debian Utilities
```
jserv, line 381: It does not make sense to preserve Debian utilities.

scripts/busybox.config
```diff
@@ -0,0 +1,1148 @@
...
+#
+CONFIG_PIPE_PROGRESS=y
+CONFIG_RUN_PARTS=y
+CONFIG_FEATURE_RUN_PARTS_LONG_OPTIONS=y
+CONFIG_FEATURE_RUN_PARTS_FANCY=y
+CONFIG_START_STOP_DAEMON=y
+CONFIG_FEATURE_START_STOP_DAEMON_LONG_OPTIONS=y
+CONFIG_FEATURE_START_STOP_DAEMON_FANCY=y
+CONFIG_WHICH=y
```
jserv, line 386: `which` can be kept though.

scripts/busybox.config
```diff
@@ -0,0 +1,1148 @@
...
+CONFIG_FEATURE_VOLUMEID_REISERFS=y
+CONFIG_FEATURE_VOLUMEID_ROMFS=y
+CONFIG_FEATURE_VOLUMEID_SQUASHFS=y
+CONFIG_FEATURE_VOLUMEID_SYSV=y
+CONFIG_FEATURE_VOLUMEID_UBIFS=y
+CONFIG_FEATURE_VOLUMEID_UDF=y
+CONFIG_FEATURE_VOLUMEID_XFS=y
+
+#
+# Miscellaneous Utilities
```
jserv, line 710: Most of Miscellaneous Utilities can be removed.

scripts/build-initrd.sh
RinHizakura: Do I misunderstood something? Am I put the command belongs to wrong function?

jserv: The problem was that this function relies on the path of busybox. You should eliminate the expectation of current path. Instead, you should add `popd` at the end of function `build_busybox`. Alternatively, use `make -C ${OUT}/busybox-${BUSYBOX_VER} all install`

scripts/build-initrd.sh
```diff
@@ -0,0 +1,40 @@
...
+   pushd ${OUT}/busybox-${BUSYBOX_VER}
+   make -j `nproc` && make install
+}
+
+function generate_rootfs()
+{
+   cd _install
+   mv linuxrc init
+   mkdir -p etc/init.d
+   cd etc/init.d
```
jserv: The command `cd` is considered harmful because you can not guarantee its success.

RinHizakura:
The script to build busybox is changed to avoid using `cd`, please check if there's still some bad design. The configs of busybox are also changed according to those comments.

scripts/build-initrd.sh
```diff
@@ -0,0 +1,37 @@
...
+    curl -L -o ${OUT}/${BUSYBOX_ARCHIVE} ${BUSYBOX_DL} || exit 1
+    echo "Extracting..."
+    tar -xjf ${OUT}/${BUSYBOX_ARCHIVE} -C ${OUT} || exit 1
+}
+
+function build_busybox()
+{
+   echo "Building Busybox..."
+   cp -f scripts/busybox.config ${OUT}/busybox-${BUSYBOX_VER}/.config
+   make -j `nproc` -C ${OUT}/busybox-${BUSYBOX_VER} all install
```
jserv: Append `|| exit 1` at the end of this command, so that it can raise errors earlier.

scripts/build-initrd.sh
```diff
@@ -0,0 +1,37 @@
...
+}
+
+function generate_rootfs()
```
jserv, line 24: Print some messages here, such as "Generating root file system..."

scripts/build-initrd.sh
```diff
@@ -0,0 +1,37 @@
...
+{
+   pushd ${OUT}/busybox-${BUSYBOX_VER}/_install
+   mv linuxrc init
+   mkdir -p etc/init.d
+   echo -e "#!/bin/sh\nmount -t devtmpfs devtmpfs /dev\nmkdir -p /dev/pts\nmount -vt devpts -o gid=4,mode=620 none /dev/pts" > etc/init.d/rcS
```
jserv: I think it would be better if we move `rcS` in directory `scripts`.

scripts/build-initrd.sh
```diff
@@ -0,0 +1,37 @@
...
+   chmod 755 etc/init.d/rcS
+   find . | cpio -o --format=newc > ../../rootfs.cpio
```
jserv: Ditto. Append `|| exit 1` at the end of the command.

scripts/busybox.config
```diff
@@ -0,0 +1,1099 @@
...
+CONFIG_GZIP=y
+CONFIG_FEATURE_GZIP_LONG_OPTIONS=y
+CONFIG_GZIP_FAST=0
+# CONFIG_FEATURE_GZIP_LEVELS is not set
+CONFIG_FEATURE_GZIP_DECOMPRESS=y
+CONFIG_LZOP=y
+# CONFIG_UNLZOP is not set
+# CONFIG_LZOPCAT is not set
+# CONFIG_LZOP_COMPR_HIGH is not set
+CONFIG_RPM=y
```
jserv: Remove `rpm` and `rpm2cpio`.

scripts/busybox.config
```diff
@@ -0,0 +1,1099 @@
...
+CONFIG_FEATURE_DATE_COMPAT=y
+CONFIG_DD=y
+CONFIG_FEATURE_DD_SIGNAL_HANDLING=y
+CONFIG_FEATURE_DD_THIRD_STATUS_LINE=y
+CONFIG_FEATURE_DD_IBS_OBS=y
+CONFIG_FEATURE_DD_STATUS=y
+CONFIG_DF=y
+CONFIG_FEATURE_DF_FANCY=y
+CONFIG_DIRNAME=y
+CONFIG_DOS2UNIX=y
```
jserv: Remove `dos2unix` and `unix2dos` since we always transfer files in UNIX format.

scripts/busybox.config
```diff
@@ -0,0 +1,1099 @@
...
+CONFIG_UNIX2DOS=y
+CONFIG_DU=y
+CONFIG_FEATURE_DU_DEFAULT_BLOCKSIZE_1K=y
+CONFIG_ECHO=y
+CONFIG_FEATURE_FANCY_ECHO=y
+CONFIG_ENV=y
+CONFIG_EXPAND=y
```
jserv: Remove `expand` and `unexpand` since we have `vi` as the editor.

scripts/busybox.config
```diff
@@ -0,0 +1,1099 @@
...
+# Editors
+#
+CONFIG_AWK=y
+CONFIG_FEATURE_AWK_LIBM=y
+CONFIG_FEATURE_AWK_GNU_EXTENSIONS=y
+CONFIG_CMP=y
+CONFIG_DIFF=y
+CONFIG_FEATURE_DIFF_LONG_OPTIONS=y
+CONFIG_FEATURE_DIFF_DIR=y
+CONFIG_ED=y
```
jserv: Remove `ed` because it is old-fashioned.

scripts/busybox.config
```diff
@@ -0,0 +1,1099 @@
...
+CONFIG_FEATURE_XARGS_SUPPORT_TERMOPT=y
+CONFIG_FEATURE_XARGS_SUPPORT_ZERO_TERM=y
+CONFIG_FEATURE_XARGS_SUPPORT_REPL_STR=y
+CONFIG_FEATURE_XARGS_SUPPORT_PARALLEL=y
+CONFIG_FEATURE_XARGS_SUPPORT_ARGS_FILE=y
+
+#
+# Init Utilities
+#
+CONFIG_BOOTCHARTD=y
```
jserv: `bootchartd` is useless now because we have no way to get its results.

scripts/busybox.config
```diff
@@ -0,0 +1,1099 @@
...
+CONFIG_BLOCKDEV=y
+CONFIG_CAL=y
+CONFIG_CHRT=y
+CONFIG_DMESG=y
+CONFIG_FEATURE_DMESG_PRETTY=y
+CONFIG_EJECT=y
+CONFIG_FEATURE_EJECT_SCSI=y
+CONFIG_FALLOCATE=y
+CONFIG_FATATTR=y
+CONFIG_FBSET=y
```
jserv: Remove `fbset` because we unlikely have framebuffer implementations.

scripts/busybox.config
```diff
@@ -0,0 +1,1099 @@
...
+CONFIG_FEATURE_ACPID_COMPAT=y
+CONFIG_BLKDISCARD=y
+CONFIG_BLKID=y
+CONFIG_FEATURE_BLKID_TYPE=y
+CONFIG_BLOCKDEV=y
+CONFIG_CAL=y
+CONFIG_CHRT=y
+CONFIG_DMESG=y
+CONFIG_FEATURE_DMESG_PRETTY=y
+CONFIG_EJECT=y
```
jserv: Remove `eject` since it never works for KVM.

scripts/busybox.config
```diff
@@ -0,0 +1,1099 @@
...
+CONFIG_FEATURE_EJECT_SCSI=y
+CONFIG_FALLOCATE=y
+CONFIG_FATATTR=y
+CONFIG_FBSET=y
+CONFIG_FEATURE_FBSET_FANCY=y
+CONFIG_FEATURE_FBSET_READMODE=y
+CONFIG_FDFORMAT=y
+CONFIG_FDISK=y
```
jserv: Both `fdformat` and `fdisk` are useless now.

scripts/tiny.config
```diff
@@ -524,7 +536,8 @@ CONFIG_HAVE_PCI=y
 # Generic Driver Options
 #
 # CONFIG_UEVENT_HELPER is not set
-# CONFIG_DEVTMPFS is not set
+CONFIG_DEVTMPFS=y
+CONFIG_DEVTMPFS_MOUNT=y
```
RinHizakura, line 540: Note: These configs are also enabled so we don't have to change the config again to boot Linux successfully in the future.

RinHizakura:
Any other things are fixed according to the comments, and they are required to be checked again, thanks!

scripts/build-linux-image.sh
```diff
@@ -26,5 +26,5 @@ function build_linux()
    popd
 }
 
-download_linux
+#download_linux
```
jserv: Why?

scripts/rcS
```diff
@@ -0,0 +1,4 @@
+#!/bin/sh
```
jserv: Someone might do `git clone` this repository on macOS (like me), and macOS file systems generally are case-insensitive. Therefore, file "rcS" might confuse the file system. The best way is to rename it to something like `rc-startup` (in git repository) and let the script to copy it into `rcS`.

src/vm.c
```diff
@@ -152,24 +141,59 @@ int vm_load(vm_t *v, const char *image_path)
...
     boot->hdr.ext_loader_ver = 0x0;
     boot->hdr.cmd_line_ptr = 0x20000;
     memset(cmdline, 0, boot->hdr.cmdline_size);
     memcpy(cmdline, KERNEL_OPTS, sizeof(KERNEL_OPTS));
     memmove(kernel, (char *) data + setupsz, datasz - setupsz);
+
+    /* setup E820 memory map to report usable memory address ranges for initrd */
+    unsigned int idx = 0;
+    boot->e820_table[idx++] = (struct boot_e820_entry){
+        .addr = 0x0, .size = ISA_START_ADDRESS - 1, .type = E820_RAM};
```
jserv: Add comma (`,`) right after `E820_RAM`, so that `clang-format` can indent better.

src/vm.c
```diff
@@ -152,24 +141,59 @@ int vm_load(vm_t *v, const char *image_path)
...
+    boot->e820_table[idx++] =
+        (struct boot_e820_entry){.addr = ISA_END_ADDRESS,
+                                 .size = RAM_SIZE - ISA_END_ADDRESS,
+                                 .type = E820_RAM};
```
jserv: Ditto.

scripts/build-linux-image.sh
RinHizakura: It was my fault. I'll fix it.

PR #3, opened 2021-09-23
Emulate the basic requiremnt of UART to boot Linux

RinHizakura:
This patch implements more emulation of the serial UART. Therefore, we can perform serial communications more completely, including receive information and interact with the command line for some BusyBox Unix utilities.

There are some options for implementation which could be discussed according to your decision of the development of this project:
* The implementation of UART now could be more complete or simpler. It depends on what we are going to do with the loading kernel image.
*  You mentioned the possibility to use coroutine instead of thread to implement UART. However, I'm not sure if it is worth for our current requirements:
    * If no condition (e.g. interrupted by I/O) which should be handled by our program happens, VM will always run inside the loop of `KVM_RUN` ioctl. If we want to yield the control back from `KVM_RUN` ioctl for the coroutine, we may need more mechanisms (e.g. timer?) to achieve this. So things would become harder.
    * The task of the created thread is very simple now. If there are no other tasks (e.g. other hardware emulation), I'm not sure of the necessity of the coroutine. Coupling with there are many ways to design for coroutine mechanism, I can't decide for the best implementation for this project.

### DIFF
--- a/src/serial.c
+++ b/src/serial.c
@@ -1,0 +1,223 @@
+#include <linux/serial_reg.h>
+#include <poll.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdio.h>
+#include <stdlib.h>
+#include <unistd.h>
+
+#include "serial.h"
+#include "utils.h"
+#include "vm.h"
+
+#define SERIAL_IRQ 4
+#define IO_READ8(data) *((uint8_t *) data)
+#define IO_WRITE8(data, value) ((uint8_t *) data)[0] = value
+
+/* global state to stop the loop of thread */
+static volatile bool thread_stop = false;
+
+struct serial_dev_priv {
+    uint8_t dll;
+    uint8_t dlm;
+    uint8_t iir;
+    uint8_t ier;
+    uint8_t fcr;
+    uint8_t lcr;
+    uint8_t mcr;
+    uint8_t lsr;
+    uint8_t msr;
+    uint8_t scr;
+
+    struct fifo rx_buf;
+};
+
+static struct serial_dev_priv serial_dev_priv = {
+    .iir = UART_IIR_NO_INT,
+    .mcr = UART_MCR_OUT2,
+    .lsr = UART_LSR_TEMT | UART_LSR_THRE,
+    .msr = UART_MSR_DCD | UART_MSR_DSR | UART_MSR_CTS,
+};
+
+/* FIXME: This implementation is incomplete */
+static void serial_update_irq(serial_dev_t *s)
+{
+    struct serial_dev_priv *priv = (struct serial_dev_priv *) s->priv;
+    uint8_t iir = UART_IIR_NO_INT;
+
+    /* If enable receiver data interrupt and receiver data ready */
+    if ((priv->ier & UART_IER_RDI) && (priv->lsr & UART_LSR_DR))
+        iir = UART_IIR_RDI;
+    /* If enable transmiter data interrupt and transmiter empty */
+    else if ((priv->ier & UART_IER_THRI) && (priv->lsr & UART_LSR_TEMT))
+        iir = UART_IIR_THRI;
+
+    priv->iir = iir | 0xc0;
+
+    /* FIXME: the return error of vm_irq_line should be handled */
+    vm_irq_line(container_of(s, vm_t, serial), SERIAL_IRQ,
+                iir == UART_IIR_NO_INT ? 0 /* inactive */ : 1 /* active */);
+}
+
+static int serial_readable(serial_dev_t *s)
+{
+    struct pollfd pollfd = (struct pollfd){
+        .fd = s->infd,
+        .events = POLLIN,
+    };
+    return (poll(&pollfd, 1, 0) > 0) && (pollfd.revents & POLLIN);
+}
+
+static void serial_console(serial_dev_t *s)
+{
+    struct serial_dev_priv *priv = (struct serial_dev_priv *) s->priv;
+
+    while (!__atomic_load_n(&thread_stop, __ATOMIC_RELAXED)) {
+        pthread_mutex_lock(&s->lock);
+
+        if (priv->lsr & UART_LSR_DR || !fifo_is_empty(&priv->rx_buf))
+            goto unlock;
+
+        while (!fifo_is_full(&priv->rx_buf) && serial_readable(s)) {
+            char c;
+            if (read(s->infd, &c, 1) == -1)
+                break;
+            if (!fifo_put(&priv->rx_buf, c))
+                break;
+            priv->lsr |= UART_LSR_DR;
+        }
+        serial_update_irq(s);
+    unlock:
+        pthread_mutex_unlock(&s->lock);
+    }
+}
+
+static void serial_in(serial_dev_t *s, uint16_t offset, void *data)
+{
+    struct serial_dev_priv *priv = (struct serial_dev_priv *) s->priv;
+
+    pthread_mutex_lock(&s->lock);
+
+    switch (offset) {
+    case UART_RX:
+        if (priv->lcr & UART_LCR_DLAB) {
+            IO_WRITE8(data, priv->dll);
+        } else {
+            if (fifo_is_empty(&priv->rx_buf))
+                break;
+
+            uint8_t value;
+            if (fifo_get(&priv->rx_buf, value))
+                IO_WRITE8(data, value);
+
+            if (fifo_is_empty(&priv->rx_buf)) {
+                priv->lsr &= ~UART_LSR_DR;
+                serial_update_irq(s);
+            }
+        }
+        break;
+    case UART_IER:
+        if (priv->lcr & UART_LCR_DLAB)
+            IO_WRITE8(data, priv->dlm);
+        else
+            IO_WRITE8(data, priv->ier);
+        break;
+    case UART_IIR:
+        IO_WRITE8(data, priv->iir | 0xc0);  // 0xc0 stands for FIFO enabled
+        break;
+    case UART_LCR:
+        IO_WRITE8(data, priv->lcr);
+        break;
+    case UART_MCR:
+        IO_WRITE8(data, priv->mcr);
+        break;
+    case UART_LSR:
+        IO_WRITE8(data, priv->lsr);
+        break;
+    case UART_MSR:
+        IO_WRITE8(data, priv->msr);
+        break;
+    case UART_SCR:
+        IO_WRITE8(data, priv->scr);
+        break;
+    default:
+        break;
+    }
+    pthread_mutex_unlock(&s->lock);
+}
+
+static void serial_out(serial_dev_t *s, uint16_t offset, void *data)
+{
+    struct serial_dev_priv *priv = (struct serial_dev_priv *) s->priv;
+
+    pthread_mutex_lock(&s->lock);
+
+    switch (offset) {
+    case UART_TX:
+        if (priv->lcr & UART_LCR_DLAB) {
+            priv->dll = IO_READ8(data);
+        } else {
+            priv->lsr |= (UART_LSR_TEMT | UART_LSR_THRE);  // flush TX
+            putchar(((char *) data)[0]);
+            fflush(stdout);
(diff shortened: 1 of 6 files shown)


src/serial.c
```diff
@@ -0,0 +1,220 @@
...
+    if ((s->ier & UART_IER_RDI) && (s->lsr & UART_LSR_DR))
+        iir = UART_IIR_RDI;
+    /* If enable transmiter data interrupt and transmiter empty */
+    else if ((s->ier & UART_IER_THRI) && (s->lsr & UART_LSR_TEMT))
+        iir = UART_IIR_THRI;
+
+    s->iir = iir | 0xc0;
+
+    /* FIXME: the return error of vm_irq_line should be handled */
+    if (iir == UART_IIR_NO_INT)
```
jserv: Rewrite as following:
```c
    vm_irq_line(container_of(s, vm_t, serial), SERIAL_IRQ,
                iir == UART_IIR_NO_INT ? 0 /* inactive */ : 1 /* active */);
```

src/serial.h
```diff
@@ -0,0 +1,43 @@
...
+
+#define COM1_PORT_BASE 0x03f8
+#define COM1_PORT_SIZE 8
+#define COM1_PORT_END (COM1_PORT_BASE + COM1_PORT_SIZE)
+
+#define FIFO_LEN 64
+
+typedef struct serial_dev serial_dev_t;
+struct serial_dev {
+    uint8_t dll;
```
jserv: Can you replace these `uint8_t` objects with single `priv`? That is, the member `void *priv` which points to the real storage for serial implementation. Thus, we can encapsulate implementation details in public headers.

src/vm.c
```diff
@@ -231,8 +230,24 @@ int vm_run(vm_t *v)
     }
 }
 
+int vm_irq_line(vm_t *v, int irq, int level)
+{
+    struct kvm_irq_level irq_level = {
+        {
+            .irq = irq,
```
jserv: Remove the trailing comma (`,`) would make fewer lines with `clang-format`.

src/serial.c
```diff
@@ -0,0 +1,236 @@
...
+}
+
+void serial_handle(serial_dev_t *s,
+                   int direction,
+                   uint16_t port,
+                   uint8_t size,
+                   uint32_t count,
+                   void *data)
+{
+    bool is_write = direction == KVM_EXIT_IO_OUT;
```
jserv: Rewrite as following:
```c
const bool is_write = (direction == KVM_EXIT_IO_OUT);
```

src/serial.c
```diff
@@ -0,0 +1,236 @@
...
+        break;
+    default:
+        break;
+    }
+    pthread_mutex_unlock(&s->mutex);
+}
+
+static void serial_out(serial_dev_t *s, uint16_t offset, void *data)
+{
+    struct serial_dev_priv *s_priv = (struct serial_dev_priv *) s->priv;
```
jserv: You can safely name it `priv` instead of weird `s_priv`.

src/serial.h
```diff
@@ -0,0 +1,30 @@
...
+#define COM1_PORT_SIZE 8
+#define COM1_PORT_END (COM1_PORT_BASE + COM1_PORT_SIZE)
+
+#define FIFO_LEN 64
+
+typedef struct serial_dev serial_dev_t;
+
+struct serial_dev {
+    void *priv;
+    pthread_mutex_t mutex;
```
jserv: Rename from `mutex` to `lock` because we might re-implement with something like coroutines later.

src/serial.c
```diff
@@ -0,0 +1,236 @@
...
+#include <unistd.h>
+
+#include "serial.h"
+#include "vm.h"
+
+#define SERIAL_IRQ 4
+#define IO_READ8(data) *((uint8_t *) data)
+#define IO_WRITE8(data, value) ((uint8_t *) data)[0] = value
+
+#define container_of(ptr, type, member)               \
```
jserv: How about moving macro `container_of` into `src/util.h`?

src/serial.c
```diff
@@ -0,0 +1,236 @@
...
+    uint8_t iir;
+    uint8_t ier;
+    uint8_t fcr;
+    uint8_t lcr;
+    uint8_t mcr;
+    uint8_t lsr;
+    uint8_t msr;
+    uint8_t scr;
+
+    uint8_t rx[FIFO_LEN];
```
jserv: Ditto. We can create a straightforward FIFO implementation and place it in `src/util.h`.

src/serial.c
```diff
@@ -0,0 +1,231 @@
...
+
+void serial_init(serial_dev_t *s)
+{
+    *s = (serial_dev_t){
+        .priv = (void *) &serial_dev_priv,
+    };
+
+    pthread_mutex_init(&s->lock, NULL);
+    s->infd = STDIN_FILENO;
+    // create a thread for waiting input
```
jserv: Change the comment to
> create a thread which accepts serial input

RinHizakura:
There's a simple FIFO implementation now. Please tell me if you want a more general/safe one.

src/serial.c
```diff
@@ -0,0 +1,230 @@
...
+}
+
+void serial_handle(serial_dev_t *s,
+                   int direction,
+                   uint16_t port,
+                   uint8_t size,
+                   uint32_t count,
+                   void *data)
+{
+    bool is_write = (direction == KVM_EXIT_IO_OUT);
```
jserv: Rewrite function body as following:
```c
    void (*serial_op)(serial_dev_t *, uint16_t, void *) = (direction == KVM_EXIT_IO_OUT) ? serial_out : serial_in;
    for (uint16_t offset = port - COM1_PORT_BASE; count--; data += size)
        serial_op(s, offset, data);
```

src/serial.c
```diff
@@ -0,0 +1,230 @@
...
+                iir == UART_IIR_NO_INT ? 0 /* inactive */ : 1 /* active */);
+}
+
+static int serial_readable(serial_dev_t *s)
+{
+    static struct timeval timeout = {.tv_sec = 0, .tv_usec = 0};
+    fd_set readfds;
+    FD_ZERO(&readfds);
+    FD_SET(s->infd, &readfds);
+    int result = select(s->infd + 1, &readfds, NULL, NULL, &timeout);
```
jserv: Why not using `poll` system call instead?

src/serial.c
```diff
@@ -0,0 +1,230 @@
...
+        if (is_write)
+            serial_out(s, offset, data);
+        else
+            serial_in(s, offset, data);
+
+        data += size;
+    }
+}
+
+void serial_exit(serial_dev_t *s)
```
jserv, line 218: Would you release the resources here?

RinHizakura: I think we should reclaim mutex here. The resource of the thread (from `pthread_create`) should be released by the way?

RinHizakura: The resource of mutex should be released here. I think I should also reclaim the resource of thread by the way?

jserv: I was worried about the thread created in function `serial_init`. Maybe the invocation of `pthread_cancel` can help from the perspective of resource management.

jserv:
> There's a simple FIFO implementation now. Please tell me if you want a more general/safe one.

Thanks. It is straightforward and reasonably functioned.

src/serial.c
RinHizakura: I'm not sure if `pthread_cancel` is a good solution because we may need a cancellation point in our thread (?). How about we just maintain a global state and loop according to it, instead of using infinite loop for our console thread. Then, if `serial_exit` is executed, the main thread (or maybe the signal handler in the future because we usually use `Ctrl+C` to stop our program) can change the state and let the console thread exit itself, so we can `pthread_join` it.

RinHizakura: We can use atomic operation to maintain the global state, so I think this could be an easier solution?

jserv: Affirmative. Let's try global state tracking with atomics. Also, `atexit` and signal handler may be used for such scenario.

RinHizakura:
I use simple counter to early stop the `vm_run` and make sure that we reclaim all resource by valgrind, if we do call `serial_exit`.

src/serial.c
```diff
@@ -0,0 +1,228 @@
...
+                iir == UART_IIR_NO_INT ? 0 /* inactive */ : 1 /* active */);
+}
+
+static int serial_readable(serial_dev_t *s)
+{
+    struct pollfd pollfd = (struct pollfd){
+        .fd = s->infd,
+        .events = POLLIN,
+    };
+    int result;
```
jserv: Rewrite as following:
```c
    return (poll(&pollfd, 1, 0) > 0) && (pollfd.revents & POLLIN);
```

src/utils.h
```diff
@@ -0,0 +1,45 @@
...
+};
+
+#define fifo_is_empty(fifo) ((fifo)->head == (fifo)->tail)
+
+#define fifo_is_full(fifo) ((fifo)->tail - (fifo)->head > FIFO_MASK)
+
+#define fifo_put(fifo, value)                               \
+    ({                                                      \
+        unsigned int __ret;                                 \
+        __ret = !fifo_is_full(fifo);                        \
```
jserv: Rewrite as following:
```c
unsigned int __ret = !fifo_is_full(fifo);
```

src/utils.h
```diff
@@ -0,0 +1,45 @@
...
+            (fifo)->data[(fifo)->tail & FIFO_MASK] = value; \
+            (fifo)->tail++;                                 \
+        }                                                   \
+        __ret;                                              \
+    })
+
+#define fifo_get(fifo, value)                               \
+    ({                                                      \
+        unsigned int __ret;                                 \
+        __ret = !fifo_is_empty(fifo);                       \
```
jserv: Ditto. Shorten.

src/serial.c
```diff
@@ -0,0 +1,228 @@
...
+#include "serial.h"
+#include "utils.h"
+#include "vm.h"
+
+#define SERIAL_IRQ 4
+#define IO_READ8(data) *((uint8_t *) data)
+#define IO_WRITE8(data, value) ((uint8_t *) data)[0] = value
+
+/* global state to stop the loop of thread */
+static volatile int thread_stop = 0;
```
jserv: Use `bool` type.

src/serial.c
```diff
@@ -0,0 +1,228 @@
...
+    void (*serial_op)(serial_dev_t *, uint16_t, void *) =
+        (direction == KVM_EXIT_IO_OUT) ? serial_out : serial_in;
+    for (uint16_t offset = port - COM1_PORT_BASE; count--; data += size)
+        serial_op(s, offset, data);
+}
+
+void serial_exit(serial_dev_t *s)
+{
+    __atomic_store_n(&thread_stop, 1, __ATOMIC_RELAXED);
+    pthread_join(s->console_tid, NULL);
```
jserv: `console_tid` sounds weird. Did you mean `worker`?

src/vm.c
```diff
@@ -213,13 +215,10 @@ int vm_run(vm_t *v)
...
         case KVM_EXIT_IO:
-            if (run->io.port == 0x3f8 && run->io.direction == KVM_EXIT_IO_OUT) {
-                uint32_t size = run->io.size;
-                uint64_t offset = run->io.data_offset;
-                printf("%.*s", size * run->io.count, (char *) run + offset);
-            } else if ((run->io.port == 0x3f8 + 5) &&
-                       (run->io.direction == KVM_EXIT_IO_IN))
-                *((char *) run + run->io.data_offset) = 0x20;
+            if (run->io.port >= COM1_PORT_BASE && run->io.port < COM1_PORT_END)
+                serial_handle(&v->serial, run->io.direction, run->io.port,
```
jserv: If function prototype of `serial_handle` is changed to two parameters, we can avoid the data copy overhead. That is,
```
void serial_handle(serial_dev_t *s, struct kvm_run *r);
```

Eventually, the object of `struct kvm_run` will be updated including its internal data.

jserv:
Thank @RinHizakura again for the great work!

Issue #4, opened 2022-02-04
Re-implement 8250 serial device


jserv:
[tinybox](https://github.com/DaiZhiyuan/tinybox) is a neat tool for hosting KVM guests. Its [8250 serial implementation](https://github.com/DaiZhiyuan/tinybox/blob/master/8250-serial.c) uses [poll(2)](https://man7.org/linux/man-pages/man2/poll.2.html) along with `kvm__irq_line` for efficient handling. There is no POSIX thread needed. We might rework the existing 8250 serial in `kvm-host`.

RinHizakura:
I just reviewed the [8250 serial implementation](https://github.com/DaiZhiyuan/tinybox/blob/master/8250-serial.c) of `tinybox`, and here's some discussion for it. 

The problem of serial implementation is: If no condition which should be handled happens, VM will always run inside the loop of `KVM_RUN` ioctl. After a little time of trying, it looks like keyboard input is useless to interrupt `KVM_RUN` ioctl.

To solve the issue, the previous solution of `kvm-host` is using another thread to handle serial input. The solution in `tinybox`, on the other hand, is using a timer to periodically return from `KVM_RUN` ioctl, just like another approach that I have mentioned in #3. I don't think the latter solution is certainly more efficient. For example, although no POSIX thread is needed for the `tinybox` solution, the timer interrupt may cause extra cost if we return from `KVM_RUN` ioctl but there's nothing that needs to be handled.

But if we only consider the simplicity, yes, the `tinybox`'s solution seems to be indeed better. So I'll still change the implementation first.

jserv:
As Maeda says, "[When in doubt, just remove. But be careful of what you remove.](https://lightsout.studio/design/the-art-of-mastering-simplicity-in-design/)"
Using single-threaded model with I/O multiplexor should be enough to implement efficient 8250 serial device. Regarding the pthread-based interrupt handling, I think we can leave it as further work.

PR #5, opened 2022-02-04
Handle serial input using timerfd mechanism 

RinHizakura:
This PR is related to #4.

The problem of serial implementation is: If no condition which should be handled happens, VM will always run inside the loop of KVM_RUN ioctl. To solve the issue, the previous solution of kvm-host is using another thread to handle serial input. On the other hand, another solution is using a timer to periodically return from KVM_RUN ioctl. The latter solution may not certainly be more efficient. For example, although no POSIX thread is needed for the latter solution, the timer interrupt may cause extra cost if we return from KVM_RUN ioctl without anything that needs to be handled.

But if we only consider the simplicity, the latter solution seems to be indeed better. So we change the implementation for such reason now.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -5,16 +5,20 @@
 #include <asm/bootparam.h>
 #include <asm/e820.h>
 
+#include <assert.h>
+#include <fcntl.h>
 #include <linux/kvm.h>
 #include <linux/kvm_para.h>
-
-#include <fcntl.h>
+#include <signal.h>
+#include <stdbool.h>
 #include <stdint.h>
 #include <stdlib.h>
 #include <string.h>
+#include <sys/epoll.h>
 #include <sys/ioctl.h>
 #include <sys/mman.h>
 #include <sys/stat.h>
+#include <sys/timerfd.h>
 #include <unistd.h>
 
 #include "err.h"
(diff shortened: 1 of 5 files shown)


src/vm.c
```diff
@@ -72,6 +73,53 @@ static void vm_init_cpu_id(vm_t *v)
...
+    timer_t timerid;
+    struct sigevent sev;
+    struct itimerspec its;
+    struct sigaction sa;
+    sigset_t mask;
+
+    sa.sa_flags = SA_SIGINFO;
+    sa.sa_sigaction = handler;
+    sigemptyset(&sa.sa_mask);
+    if (sigaction(SIGALRM, &sa, NULL) == -1)
```
jserv: How about using timerfd instead of `SIGALRM`?

RinHizakura: The advantage of `SIGALRM` is the signal to interrupt periodically, which can depend on the standard library / operating system mechanism directly without extra codes. If we want to use timerfd instead, we may need a thread to poll the file descriptor for whether it is readable, and trigger interrupts explicitly if the timer expires. Otherwise, VM will always run inside the loop of KVM_RUN ioctl and never return.

RinHizakura: Except we can have some good approaches to listen for the timerfd without other threads, I think `SIGALRM` is more simple since we don't actually need to do anything in the signal handler.

jserv: Can you try the mixture of timerfd and epoll instead?

RinHizakura: Sure, maybe I can show how I think about the implementation using timerfd first. Then we can compare both methods.

RinHizakura:
Here's the implementation that uses timerfd and epoll mechanism.

jserv (changes requested):
The subject was not precise. In fact, POSIX Thread is in use.

jserv:
Sometime, Linux boots without entering `init` process. The last message:
```
Console: colour dummy device 80x25
printk: console [ttyS0] enabled
APIC: ACPI MADT or MP tables are not detected
APIC: Switch to virtual wire mode setup with no configuration
Not enabling interrupt remapping due to skipped IO-APIC setup
```

jserv:
This patch does make 8250 serial console more responsive.

src/vm.c
```diff
@@ -72,6 +76,87 @@ static void vm_init_cpu_id(vm_t *v)
...
+
+    timer_fd = timerfd_create(CLOCK_MONOTONIC, TFD_CLOEXEC | TFD_NONBLOCK);
+    if (timer_fd == -1)
+        return throw_err("Failed to create timer");
+
+    ev.events = EPOLLIN, ev.data.fd = timer_fd;
+    if (epoll_ctl(epoll_fd, EPOLL_CTL_ADD, timer_fd, &ev) == -1)
+        return throw_err("Failed to add event for timer");
+
+    its.it_value.tv_sec = FREQ_NS / 1000000000;
```
jserv: Use IEEE 754 notations such as `1.0e9` and `1.0e6` instead.

src/vm.c
```diff
@@ -72,6 +76,87 @@ static void vm_init_cpu_id(vm_t *v)
...
+    sigemptyset(&mask);
+    sigaddset(&mask, SIGUSR1);
+    if (sigprocmask(SIG_SETMASK, &mask, NULL) == -1)
+        return throw_err("Failed to block timer signal");
+
+    epoll_fd = epoll_create1(0);
+    if (epoll_fd == -1)
+        return throw_err("Failed to created epoll fd");
+
+    timer_fd = timerfd_create(CLOCK_MONOTONIC, TFD_CLOEXEC | TFD_NONBLOCK);
```
jserv, line 135: Change to `CLOCK_MONOTONIC_RAW` since we don't care about NTP.

src/vm.c
```diff
@@ -72,6 +76,87 @@ static void vm_init_cpu_id(vm_t *v)
...
+
+    sa.sa_flags = SA_SIGINFO;
+    sa.sa_sigaction = handler;
+    sigemptyset(&sa.sa_mask);
+    if (sigaction(SIGUSR1, &sa, NULL) == -1)
+        return throw_err("Failed to create signal handler");
+
+    /* Block timer signal temporarily. */
+    sigemptyset(&mask);
+    sigaddset(&mask, SIGUSR1);
```
jserv, line 127: Since `SIGUSR1` is introduced for flexible flow control, I just wonder if we can remove the use of POSIX Threads completely.

RinHizakura:
> Sometime, Linux boots without entering `init` process. The last message:
> 
> ```
> Console: colour dummy device 80x25
> printk: console [ttyS0] enabled
> APIC: ACPI MADT or MP tables are not detected
> APIC: Switch to virtual wire mode setup with no configuration
> Not enabling interrupt remapping due to skipped IO-APIC setup
> ```

I'll meet this problem sometimes too. I'm not sure for the reason to this.

src/vm.c
RinHizakura: `CLOCK_MONOTONIC_RAW` is the invalid value for timerfd, see [fs/timerfd.c](https://elixir.bootlin.com/linux/latest/source/fs/timerfd.c#L406)

src/vm.c
RinHizakura: The key point for the implementation of the serial input feature is: to interrupt the main thread. This is not simple to achieve without thread. But maybe we can interrupt using a mechanism other than signal if there's any better solution.

Taking [kvmtool](https://github.com/kvmtool/kvmtool/blob/master/term.c) for example, it also creates a POSIX thread. But instead of listening at the timerfd, it listens at the STDIN or something similarly directly (I think this sounds like a better idea! How about changing to this design?).  And it looks like to interrupt the main thread with virtio interface instead of using signal(I'm not sure for this part).

jserv: Did you mean the way to use `epoll` to listen for STDIN events? It sounds straightforward and efficient.

RinHizakura: Yes, instead of listening to the timerfd to interrupt the main thread periodically, `kvmtool` listens to STDIN events and interrupts the main thread only when there do really have keyboard inputs. I think the latter implementation is better and asking if we want to change for that.

RinHizakura: Although we still need a thread for this implementation, now the thread doesn't access `vm_t`. It just tells the main thread that it needs to spend some time for the serial inputs first. So we don't need to worry about data race under this design.

RinHizakura: The advantage of using `epoll` to listen for STDIN events:
* Compare to the design of kvm-host on current master branch, we can remove the mutex
* Compare to the design of my original pull request, we can reduce the useless interrupt on the main thread

jserv: Affirmative! Please work on listening for STDIN events. Once you are about to revise the implementation, change the subject of this pull request.

RinHizakura:
Due to the large changes for the entire design, I re-issued the PR to #7

PR #6, opened 2022-02-08
Unmap the unused memory mappings

u1f383:
There are some memory mappings need to be unmapped after use


RinHizakura (approved):
LGTM, thanks!

jserv:
Thank @u1f383 for contributing!

PR #7, opened 2022-02-10
Handle serial input without lock

RinHizakura:
This PR is going to solved #4, and some of the discussion and where this comes from can be found in #5.

The problem of serial implementation is: If no specific condition which should be handled in userspace happens, VM will always run inside the loop of `KVM_RUN` ioctl in kernel. The user's keyboard inputs can't make it return from kernel directly.

To solve the issue, the previous solution of `kvm-host` uses another worker thread to handle serial input itself, which will need an extra mutex because it will access the VM-related data. However, under the implementation, we can't manage the code
flow of main thread and worker thread narrowly, since they'll schedule by our operating system. And we also need to carefully define the critical section for both the performance and correctness.

Because of those reasons, this patch makes a reimplementation. Now, although we still need a thread
for `kvm-host`, the worker thread doesn't access the VM-related data. Instead, it just "tells" the main thread that it needs to spend some time for the serial inputs first by signal. So we don't need to worry about data race under this design, which makes everything simple. And we can also improve the cooperative because now we can decide the code flow between serial input and KVM instead of partially depending on the operating system.

### DIFF
--- a/src/serial.c
+++ b/src/serial.c
@@ -1,5 +1,6 @@
 #include <linux/serial_reg.h>
 #include <poll.h>
+#include <signal.h>
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
@@ -7,6 +8,7 @@
 #include <stdlib.h>
 #include <unistd.h>
 
+#include "err.h"
 #include "serial.h"
 #include "utils.h"
 #include "vm.h"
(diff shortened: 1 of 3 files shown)


RinHizakura:
Note that I use `poll` instead of `epoll` for serial input because it makes us reuse the previous codes.

jserv (changes requested):
Change the subject to "Handle serial inputs without locks"

jserv:
Thank @RinHizakura for the great work!

PR #8, opened 2022-02-14
Disable canonical mode on host

RinHizakura:
The commit configures the standard input in raw mode, so we can pass unmodified user input to the guest operating system.

### DIFF
--- a/src/main.c
+++ b/src/main.c
@@ -1,5 +1,7 @@
 #include <getopt.h>
 #include <stdlib.h>
+#include <termios.h>
+#include <unistd.h>
 
 #include "err.h"
 #include "vm.h"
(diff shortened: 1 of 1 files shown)


src/main.c
```diff
@@ -17,6 +19,32 @@ static void usage(const char *execpath)
...
+static void reset_input_mode(void)
+{
+    tcsetattr(STDIN_FILENO, TCSANOW, &saved_attributes);
+}
+
+static void set_input_mode(void)
+{
+    struct termios tattr;
+    /* Make sure stdin is a terminal. */
+    if (!isatty (STDIN_FILENO))
```
jserv: Indent with `clang-format`

PR #11, opened 2022-06-21
Implement the basis of virtio-blk

ray90514:
The support of the virtio-blk device enables Guest Linux to access
the disk image in Host. It passes the test on ext4 file system.

We also introduce virtio-pci and virtq which can be utilized by other
virtio devices just like virtio-blk.


src/virtio-pci.c
```diff
@@ -0,0 +1,266 @@
...
+                           size);
+            }
+            /* guest notify buffer avail */
+            else if (offset ==
+                     offsetof(struct virtio_pci_config, notify_data)) {
+                virtq_handle_avail(&dev->vq[dev->config.notify_data.vqn]);
+            }
+            break;
+        }
+
```
jserv: You can add a `return` statement here and then the followed `else` can be eliminated.

src/virtio-pci.c
```diff
@@ -0,0 +1,266 @@
...
+{
+    struct virtio_pci_cap *caps[VIRTIO_PCI_CAP_NUM + 1];
+
+    for (int i = 1; i < VIRTIO_PCI_CAP_NUM + 1; i++) {
+        caps[i] = dev->pci_dev.hdr + next;
+        *caps[i] =
+            (struct virtio_pci_cap){.cap_vndr = PCI_CAP_ID_VNDR,
+                                    .cfg_type = i,
+                                    .cap_len = sizeof(struct virtio_pci_cap),
+                                    .bar = 0};
```
jserv: Add a comma (`,`) right after `= 0`, so that `clang-format` can indent better.

src/main.c
```diff
@@ -6,7 +6,7 @@
 #include "err.h"
 #include "vm.h"
 
-static char *kernel_file = NULL, *initrd_file = NULL;
+static char *kernel_file = NULL, *initrd_file = NULL, *disk_file = NULL;
```
jserv: For the sake of consistency, we can call it `diskimg_file`.

See https://en.wikipedia.org/wiki/Disk_image

src/virtio-pci.c
```diff
@@ -0,0 +1,266 @@
...
+        dev->config.common_cfg.device_feature = feature >> 32;
+    else
+        dev->config.common_cfg.device_feature = 0;
+}
+
+static void virtio_pci_write_guest_feature(struct virtio_pci_dev *dev)
+{
+    uint32_t select = dev->config.common_cfg.guest_feature_select;
+    uint32_t feature = dev->config.common_cfg.guest_feature;
+    if (select == 0)
```
jserv: Use switch-case instead.

configs/linux.config
```diff
@@ -1312,12 +1403,185 @@ CONFIG_INIT_STACK_NONE=y
 # end of Kernel hardening options
 # end of Security options
 
-# CONFIG_CRYPTO is not set
+CONFIG_CRYPTO=y
```
jserv, line 1406: Can you disable `CONFIG_CRYPTO`?

src/disk-image.c
```diff
@@ -0,0 +1,39 @@
+#include <fcntl.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include "disk-image.h"
```
jserv: Let's use "diskimg.[ch]" for short.

src/disk-image.c
```diff
@@ -0,0 +1,39 @@
...
+
+ssize_t disk_image_read(struct disk_image *img,
```
jserv: Ditto. All functions associated with disk image should have prefix "diskimg_".

configs/linux.config
ray90514: No, I can't. It is selected by `CONFIG_EXT4_FS` and `CONFIG_BLOCK`

src/vm.c
```diff
@@ -275,9 +285,58 @@ int vm_irq_line(vm_t *v, int irq, int level)
     return 0;
 }
 
+int vm_irq_trigger(vm_t *v, int irq)
```
jserv: When will this function be used?

src/vm.c
```diff
@@ -275,9 +285,58 @@ int vm_irq_line(vm_t *v, int irq, int level)
...
+{
+    if (vm_irq_line(v, irq, 1) < 0)
+        return -1;
+    if (vm_irq_line(v, irq, 0) < 0)
+        return -1;
+    return 0;
+}
+
+void *vm_host_to_guest(vm_t *v, void *host)
```
jserv: Ditto. Is it necessary?

src/virtio-pci.c
```diff
@@ -0,0 +1,279 @@
...
+    PCI_HDR_WRITE(dev->pci_dev.hdr, PCI_CLASS_REVISION, class << 8, 32);
+    PCI_HDR_WRITE(dev->pci_dev.hdr, PCI_INTERRUPT_LINE, irq_line, 8);
+}
+
+void virtio_pci_init(struct virtio_pci_dev *dev,
+                     struct pci *pci,
+                     struct bus *io_bus,
+                     struct bus *mmio_bus)
+{
+    uint8_t cap_list = 0x40;
```
jserv, line 259: Add comments here.

src/vm.c
ray90514: It is a replacement of irqfd and can be removed.

src/main.c
```diff
@@ -17,6 +17,8 @@ static void usage(const char *execpath)
 
     print_option("-h, --help", "Print help of CLI and exit.\n");
     print_option("-i, --initrd initrd", "Initial RAM disk image\n");
+    print_option("-d, --disk disk-image",
+                 "Initial disk image as a block device with virtio-blk\n");
```
jserv: It is not tended to be "initial disk." Precisely, it was meant to be disk image.

src/virtio-pci.c
```diff
@@ -0,0 +1,280 @@
...
+    dev->device_feature |=
+        (1ULL << VIRTIO_F_RING_PACKED) | (1ULL << VIRTIO_F_VERSION_1);
+}
+
+void virtio_pci_enable(struct virtio_pci_dev *dev)
+{
+    pci_dev_register(&dev->pci_dev);
+}
+
+void virtio_pci_exit() {}
```
jserv: Add comments which address TODO/FIXME.

jserv:
I defer to @RinHizakura .

src/diskimg.c
```diff
@@ -0,0 +1,39 @@
...
+
+#include "diskimg.h"
+
+ssize_t diskimg_read(struct diskimg *diskimg,
+                     void *data,
+                     off_t offset,
+                     size_t size)
+{
+    lseek(diskimg->fd, offset, SEEK_SET);
+    return read(diskimg->fd, data, size);
```
RinHizakura, line 13: Although it's fine to not change this currently, I'm wondering whether it will be better to use the disk image as a mmap file just like how we do on initrd or kernel image? We may have to compare two different methods in the future.

src/virtq.c
```diff
@@ -0,0 +1,66 @@
...
+    return desc;
+}
+
+void virtq_handle_avail(struct virtq *vq)
+{
+    if (!vq->info.enable)
+        return;
+    vq->ops->complete_request(vq);
+    if (vq->guest_event->flags == VRING_PACKED_EVENT_FLAG_ENABLE)
+        vq->ops->notify_used(vq);
```
RinHizakura: Replace to `virtq_complete_request` / `virtq_enable` since we have such function interface.

README.md
```diff
@@ -28,12 +28,13 @@ make check
...
 
 ```
-build/kvm-host -k bzImage [-i initrd]
+build/kvm-host -k bzImage [-i initrd] [-d disk-image]
 ```
 
 `bzImage` is the path to linux kernel bzImage. The bzImage file is in a specific format,
 containing concatenated `bootsect.o + setup.o + misc.o + piggy.o`. `initrd` is the path to
 initial RAM disk image, which is an optional argument.
+`disk-image` is the path to ext4 disk image. The disk image is as a block device with virtio-blk in Guest Linux.
```
RinHizakura: The disk image is not restricted to ext4, right? If so, I think we have to fix the description here.

By the way, can we provide either scripts or more information to build and test the disk image? Just like how we do for `bzImage` and `initrd`.

src/virtio-blk.c
```diff
@@ -0,0 +1,161 @@
...
+        v, (void *) vq->info.desc_addr);
+    vq->device_event = (struct vring_packed_desc_event *) vm_guest_to_host(
+        v, (void *) vq->info.device_addr);
+    vq->guest_event = (struct vring_packed_desc_event *) vm_guest_to_host(
+        v, (void *) vq->info.driver_addr);
+
+    uint64_t addr = virtio_pci_get_notify_addr(&dev->virtio_pci_dev, vq);
+    vm_ioeventfd_register(v, dev->ioeventfd, addr,
+                          dev->virtio_pci_dev.notify_cap->cap.length, 0);
+    pthread_create(&dev->vq_avail_thread, NULL, virtio_blk_vq_avail_handler,
```
RinHizakura, line 75: Recycle the thread resource by `pthread_join` when exiting.

README.md
ray90514: > The disk image is not restricted to ext4, right?

Yes, but we only enable the ext4 support on `linux.config`.

src/virtio-blk.c
```diff
@@ -0,0 +1,161 @@
...
+    dev->diskimg = NULL;
+    for (int i = 0; i < VIRTIO_BLK_VIRTQ_NUM; i++)
+        virtq_init(&dev->vq[i], dev, &ops);
+}
+
+void virtio_blk_exit(struct virtio_blk_dev *dev)
+{
+    if (dev->diskimg)
+        diskimg_exit(dev->diskimg);
+    virtio_pci_exit(&dev->virtio_pci_dev);
```
RinHizakura, line 194: `virtio_blk_exit` will be executed even if we don't input the disk image, but `dev->diskimg` is uninitialized at that time. We should not access it in this case.

README.md
RinHizakura: Since the user can build their Linux instead of using what we provide as an example on this project, it doesn't really matter what the config is. 

I will prefer to write something like "`disk-image` is the path to the disk image which is used as a block device with virtio-blk in Guest Linux. For example, you could use an Ext4 filesystem disk image."

src/virtq.c
```diff
@@ -0,0 +1,66 @@
...
+{
+    vq->ops->enable_vq(vq);
+}
+
+void virtq_disable(struct virtq *vq) {}
+
+#define VIRTQ_SIZE 128
+void virtq_init(struct virtq *vq, void *dev, struct virtq_ops *ops)
+{
+    vq->info.size = VIRTQ_SIZE;
```
RinHizakura, line 27: Why not take this value from `(struct virtio_pci_dev *) dev->config.common_cfg.queue_size` but using a hard code value instead?

src/virtio-blk.c
```diff
@@ -0,0 +1,161 @@
...
+
+void virtio_blk_init_pci(struct virtio_blk_dev *virtio_blk_dev,
+                         struct diskimg *diskimg,
+                         struct pci *pci,
+                         struct bus *io_bus,
+                         struct bus *mmio_bus)
+{
+    virtio_blk_init(virtio_blk_dev);
+    virtio_blk_dev->diskimg = diskimg;
+    virtio_blk_dev->config.capacity = diskimg->size >> 9;
```
RinHizakura: Initialize `diskimg` and `config.capacity` along with `virtio_blk_init`

src/virtq.c
ray90514: The value of `queue_size` depends on the specific queue selected by `queue_select`.

src/diskimg.c
ray90514: If we want to change this to mmap, notice that `virtio-blk` should support the flush command.

src/virtq.c
RinHizakura: Sorry for I might get it wrong, so will this field get modified during executing Linux in somewhere like `virtio_pci_space_write`? Or the queue size will always be assumed to be size 128 for this implementation?

src/diskimg.c
RinHizakura: Thanks! I just have a sudden thought here, but it's nice to use the `lseek` + `read`/`write` method now.

README.md
```diff
@@ -28,12 +28,13 @@ make check
...
 
 ```
-build/kvm-host -k bzImage [-i initrd]
+build/kvm-host -k bzImage [-i initrd] [-d disk-image]
 ```
 
 `bzImage` is the path to linux kernel bzImage. The bzImage file is in a specific format,
 containing concatenated `bootsect.o + setup.o + misc.o + piggy.o`. `initrd` is the path to
 initial RAM disk image, which is an optional argument.
+`disk-image` is the path to the disk image which is used as a block device with virtio-blk in Guest Linux. For example, you could use an Ext4 Filesystem disk image.
```
RinHizakura: It's my fault. Change the letter "F" in the word "Filesystem" to lowercase

src/virtq.c
ray90514: The size of a queue will not change, but the `queue_size` will be modified in `virtio_pci_select_virtq()`.

RinHizakura: I see. Sorry for misunderstanding how we sync the queue size between OS and virtio. 

No problem for this question now.

src/virtio-blk.c
```diff
@@ -0,0 +1,169 @@
...
+    dev->enable = false;
+    for (int i = 0; i < VIRTIO_BLK_VIRTQ_NUM; i++)
+        virtq_init(&dev->vq[i], dev, &ops);
+}
+
+void virtio_blk_exit(struct virtio_blk_dev *dev)
+{
+    if (!dev->enable)
+        return;
+    pthread_join(dev->vq_avail_thread, NULL);
```
RinHizakura, line 192: You may have to stop thread from the while loop like `serial.c`, otherwise the main thread will stuck at this join

src/virtio-blk.c
```diff
@@ -0,0 +1,169 @@
...
+    vm_t *v = container_of(dev, vm_t, virtio_blk_dev);
+
+    memset(dev, 0x00, sizeof(struct virtio_blk_dev));
+    dev->ioeventfd = eventfd(0, EFD_CLOEXEC);
+    dev->irqfd = eventfd(0, EFD_CLOEXEC);
+    /* FIXME: irq_num should be different to other devs */
+    dev->irq_num = 15;
+    vm_irqfd_register(v, dev->irqfd, dev->irq_num, 0);
+    dev->diskimg = NULL;
+    dev->enable = false;
```
RinHizakura: I have no idea when will this flag be set to true. Do I miss it?

ray90514: I'm sorry forget it in `virtio_blk_enable()`

RinHizakura: It looks like you change to initialize those VIRTIO structures/threads whether we give a disk image or not now, but I think what you do before my review is good enough. 

To solve the uninitialized problem, you can just add a flag variable to hint that we don't initialize `struct virtio_blk_dev virtio_blk_dev` under `vm_t` so there's no need to do `virtio_blk_exit`.  Or do you have any concerns to initialize the VIRTIO device anytime?

RinHizakura: Sometimes I will trickly return directly from `vm_run` and use valgrind to see if everything is freed correctly(don't use Ctrl+C for the normal exit flow).  You can try this yourself to make sure that :)

src/virtio-blk.c
```diff
@@ -0,0 +1,196 @@
...
+                   (void *) dev);
+}
+
+void virtio_blk_init_pci(struct virtio_blk_dev *virtio_blk_dev,
+                         struct pci *pci,
+                         struct bus *io_bus,
+                         struct bus *mmio_bus)
+{
+    virtio_blk_init(virtio_blk_dev);
+    struct virtio_pci_dev *dev = &virtio_blk_dev->virtio_pci_dev;
```
RinHizakura, line 167: How about making ` virtio_blk_init` and `virtio_blk_enable` become static,  and do `virtio_blk_enable` in `virtio_blk_init_pci` instead of seperate them in `vm_load_diskimg`?

So it will be clear that  `virtio_blk_exit` should only be run when we did  `virtio_blk_init_pci` before.

ray90514: But the value of `enable` in `virtio_blk` isn't specified if we haven't called `virtio_blk_init_pci`

RinHizakura: I see. I think what you modify in a few minutes ago is good enough for me.

RinHizakura (approved):
All the remaining looks good to me, thanks!

If @jserv doesn't have any comment then we can merge this.

README.md
```diff
@@ -28,12 +28,13 @@ make check
...
 
 ```
-build/kvm-host -k bzImage [-i initrd]
+build/kvm-host -k bzImage [-i initrd] [-d disk-image]
 ```
 
 `bzImage` is the path to linux kernel bzImage. The bzImage file is in a specific format,
 containing concatenated `bootsect.o + setup.o + misc.o + piggy.o`. `initrd` is the path to
 initial RAM disk image, which is an optional argument.
+`disk-image` is the path to the disk image which is used as a block device with virtio-blk in Guest Linux. For example, you could use an Ext4 filesystem disk image.
```
jserv: Rewrite the description to
> `disk-image` is the path to disk image which can be mounted as a block device via virtio. For the reference Linux guest, ext4 filesystem is used for disk image.

jserv:
Thank @ray90514 for contributing!

jserv:
@ray90514, you can contribute new script(s) which can create an ext4-formatted disk image as part of reference Linux guest.

PR #14, opened 2023-05-17
Use bus interface in the serial device

### LINKED ISSUE #10: Implement UART device on the same abstraction of PCI
In #9 , we have an abstraction of bus device with read and write method now. Since we have PCI and UART device emulation now, and there could be more in the future. We should integrate them with the same abstractions.

yanjiew1:
Originally, the I/O port handler checks the hard-coded I/O port range of the serial device and call the serial device I/O handler if the port number is within the range.

This pull request moves the bus interface and its implementations into dedicated files (bus.c and bus.h), changes the serial device to register its I/O port using the bus interface and removes the hard-coded port numbers in VM exit handling routine.

This fixes #10.


src/bus.h
```diff
@@ -0,0 +1,38 @@
+#pragma once
+
+#include <stdint.h>
+
+struct dev;
```
RinHizakura, line 5: Look like this can be removed.

RinHizakura (approved):
LGTM, thank you!

jserv:

PR #16, opened 2023-06-13
Handle multi-threaded FIFO operations in serial device

yanjiew1:
After this change, the I/O handler called during VM exit will only need to deal with the FIFO buffers on the device. The work to interact with the terminal is offloaded to the worker thread.

The main thread use eventfd to signal the worker thread. The worker thread uses epoll to monitor stdin, stdout and the eventfd and will be notified when the state of one of these file descriptors change and perform the I/O when necessary.

Besides changing the original implementation to use eventfd and epoll, I also make the emulated device hehave more like a physical device. The notable change is that the loopback mode is implemented.

### DIFF
--- a/src/serial.c
+++ b/src/serial.c
@@ -1,11 +1,17 @@
+#include <fcntl.h>
 #include <linux/serial_reg.h>
-#include <poll.h>
 #include <signal.h>
+#include <stdatomic.h>
 #include <stdbool.h>
 #include <stddef.h>
 #include <stdint.h>
 #include <stdio.h>
 #include <stdlib.h>
+#include <sys/epoll.h>
+#include <sys/eventfd.h>
+#include <sys/stat.h>
+#include <sys/types.h>
+#include <sys/uio.h>
 #include <unistd.h>
 
 #include "err.h"
@@ -17,7 +23,22 @@
 #define IO_READ8(data) *((uint8_t *) data)
 #define IO_WRITE8(data, value) ((uint8_t *) data)[0] = value
 
+#define IER_MASK 0x0f
+#define MCR_MASK 0x1f
+#define FCR_MASK 0xc9
+#define DEFAULT_MSR UART_MSR_DCD | UART_MSR_DSR | UART_MSR_CTS
+
+#define SERIAL_EPOLL_EVENT 1
+#define SERIAL_EPOLL_OUT 2
+#define SERIAL_EPOLL_IN 3
+
+#define SERIAL_TRIG_LEVELS \
+    {                      \
+        1, 16, 32, 56      \
+    }
+
 struct serial_dev_priv {
+    /* Device registers */
     uint8_t dll;
     uint8_t dlm;
     uint8_t iir;
@@ -28,8 +49,27 @@ struct serial_dev_priv {
     uint8_t lsr;
     uint8_t msr;
     uint8_t scr;
+    bool thr_ipending;
 
+    /* Buffers */
+    struct fifo tx_buf;
     struct fifo rx_buf;
+    uint8_t rxtrig;
+
+    /* File descriptors */
+    int infd;
+    int outfd;
+    int evfd;
+    int epollfd;
+
+    /* Worker */
+    pthread_t worker;
+    pthread_mutex_t lock;
+    pthread_mutex_t loopback_lock;
+    bool stopped;
+
+    /* Initialized */
+    bool initialized;
 };
 
 static struct serial_dev_priv serial_dev_priv = {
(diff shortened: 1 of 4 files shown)


jserv (changes requested):
Improve the git commit messages by fixing some mistakes.

jserv:
gcc-9 complains as the following:
```
src/serial.c: In function ‘serial_in’:
src/serial.c:327:9: error: a label can only be part of a statement and a declaration is not a statement
  327 |         uint8_t iir = __atomic_load_n(&priv->iir, __ATOMIC_ACQUIRE);
      |         ^~~~~~~
src/serial.c:345:9: error: a label can only be part of a statement and a declaration is not a statement
  345 |         uint8_t lsr = __atomic_load_n(&priv->lsr, __ATOMIC_RELAXED);
      |         ^~~~~~~
```

Quick fix:
```diff
--- a/src/serial.c
+++ b/src/serial.c
@@ -323,7 +323,7 @@ static void serial_in(serial_dev_t *s, uint16_t offset, void *data)
         else
             IO_WRITE8(data, priv->ier);
         break;
-    case UART_IIR:
+    case UART_IIR: {
         uint8_t iir = __atomic_load_n(&priv->iir, __ATOMIC_ACQUIRE);
         IO_WRITE8(data, iir);
         if ((iir & UART_IIR_ID) == UART_IIR_THRI) {
@@ -335,13 +335,14 @@ static void serial_in(serial_dev_t *s, uint16_t offset, void *data)
             pthread_mutex_unlock(&priv->lock);
         }
         break;
+    }
     case UART_LCR:
         IO_WRITE8(data, priv->lcr);
         break;
     case UART_MCR:
         IO_WRITE8(data, priv->mcr);
         break;
-    case UART_LSR:
+    case UART_LSR: {
         uint8_t lsr = __atomic_load_n(&priv->lsr, __ATOMIC_RELAXED);
         IO_WRITE8(data, lsr);
         if (lsr & UART_LSR_BRK_ERROR_BITS) {
@@ -352,6 +353,7 @@ static void serial_in(serial_dev_t *s, uint16_t offset, void *data)
             pthread_mutex_unlock(&priv->lock);
         }
         break;
+    }
     case UART_MSR:
         IO_WRITE8(data, priv->msr);
         if (priv->msr & UART_MSR_ANY_DELTA) {
```

src/utils.h
```diff
@@ -1,5 +1,4 @@
-#ifndef UTILS_H
-#define UTILS_H
+#pragma once
```
jserv, line 1: Don't do that. Minimize the necessary changes.

src/serial.c
```diff
@@ -28,78 +49,243 @@ struct serial_dev_priv {
...
+    int epollfd;
+
+    /* Worker */
+    pthread_t worker;
+    pthread_mutex_t lock;
+    pthread_mutex_t loopback_lock;
+    bool stopped;
+
+    /* Initialized */
+    bool initialized;
```
RinHizakura, line 72: What's the variable used for? When will we reinitialize the serial device incorrectly?

src/serial.c
```diff
@@ -210,44 +481,108 @@ static void serial_handle_io(void *owner,
...
+    struct epoll_event event;
+
+    if (priv->initialized)
+        return throw_err("Serial device is already initialized\n");
+
+    s->priv = &serial_dev_priv;
+
+    /* Create necessory file descriptors */
+    int evfd, infd, outfd, epollfd;
+    evfd = infd = outfd = epollfd = -1;
```
RinHizakura, line 496: I don't think closing a file descriptor with a negative value is a great idea for error handling (when `goto err`). It may set errno but we are not going to check it.

src/serial.c
```diff
@@ -210,44 +481,108 @@ static void serial_handle_io(void *owner,
...
+    if (epoll_ctl(epollfd, EPOLL_CTL_ADD, outfd, &event) < 0) {
+        if (errno != EPERM) {
+            throw_err("Failed to add stdout to epoll\n");
+            goto err;
+        }
+    }
+
+    /* Setup mutex */
+    pthread_mutex_init(&priv->lock, NULL);
+    pthread_mutex_init(&priv->loopback_lock, NULL);
```
RinHizakura, line 548: Do `pthread_mutex_destroy` when doing error handling and closing the serial device (`serail_exit`) will be a good practice.

src/serial.c
```diff
@@ -28,78 +49,243 @@ struct serial_dev_priv {
...
+        if (ret < 1)
+            continue;
+        if (event.data.u32 == SERIAL_EPOLL_EVENT)
+            if (read(priv->evfd, &tmp, 8) < 0)
+                throw_err("Failed to read from eventfd\n");
+        pthread_mutex_lock(&priv->loopback_lock);
+        switch (event.data.u32) {
+        case SERIAL_EPOLL_EVENT:
+            serial_transmit(s);
+            serial_receive(s);
```
RinHizakura, line 245: I don't think that you fallthrough the case intentionally. Add `break` for each case.

src/serial.c
```diff
@@ -28,78 +49,243 @@ struct serial_dev_priv {
...
+
+    /* Initialized */
+    bool initialized;
 };
 
 static struct serial_dev_priv serial_dev_priv = {
     .iir = UART_IIR_NO_INT,
     .mcr = UART_MCR_OUT2,
     .lsr = UART_LSR_TEMT | UART_LSR_THRE,
     .msr = UART_MSR_DCD | UART_MSR_DSR | UART_MSR_CTS,
```
RinHizakura, line 79: Change this to `DEFAULT_MSR`

yanjiew1:
Hello, I have a another solution to solve the problem of signal input. This solution is not based on epoll and eventfd.

Instead of using signal to notify the main thread to read from stdin, just perform the read in the worker thread. Use mutex lock to protect the serial device's internal structure and use conditional variable, so that the main thread can notify the worker thread that it should read more bytes from stdin.

Please see this commit. Instead of changing a lot of codes to not only solve the problem but also add new features, I try to make minimal changes to the code to solve the problem of serial input.
https://github.com/yanjiew1/kvm-host/commit/edb3aa4c98a0e1a94e0421f04dcf674284f8c5db

jserv:
> Please see this commit. Instead of changing a lot of codes to not only solve the problem but also add new features, I try to make minimal changes to the code to solve the problem of serial input. [yanjiew1@edb3aa4](https://github.com/yanjiew1/kvm-host/commit/edb3aa4c98a0e1a94e0421f04dcf674284f8c5db)

@RinHizakura, what do you think of the above approach?

RinHizakura:
This one looks better for me! In my opinion, it would be great to keep kvm-host as simple as possible(but still complete and efficient) as it is a "minimalist hypervisor".

jserv:
> This one looks better for me! In my opinion, it would be great to keep kvm-host as simple as possible(but still complete and efficient) as it is a "minimalist hypervisor".

Agree.
@yanjiew1, Please rebase and rework the handling of multi-threaded FIFO operations properly.

PR #17, opened 2023-06-15
Separate x86-64 specific code out of generic implementation

yanjiew1:
I successfully ported the kvm-host to arm64. These are what I modified:
- Separate architecture specific codes from architecture-neutral codes
- Import libfdt source code for arm64. (the license is compatible)
- Use dynamic IRQ number assignment rather than hard-coded IRQ number in virtio-blk and serial device
- Not registering PCI bus to io_bus and mmio_bus.  Leave this work for architecture specific initialization routine
- Fix serial input bug by reading from stdin in worker thread. Use a mutex lock and a conditional variable for synchronization
- The error reporting routine `throw_err` can support printf style formatting.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,12 @@
 include mk/common.mk
 
+ARCH ?= $(shell uname -m)
+PWD ?= $(shell pwd)
+
 CC ?= gcc
 CFLAGS = -O2
 CFLAGS += -Wall -std=gnu99
+CFLAGS += -I$(PWD)/src
 CFLAGS += -g
 LDFLAGS = -lpthread
 
@@ -21,6 +25,13 @@ OBJS := \
 	virtio-blk.o \
 	diskimg.o \
 	main.o
+
+ifeq ($(ARCH), x86_64)
+	CFLAGS += -I$(PWD)/src/arch/x86
+	CFLAGS += -include src/arch/x86/desc.h
+	OBJS += arch/x86/vm.o
+endif
+
 OBJS := $(addprefix $(OUT)/,$(OBJS))
 deps := $(OBJS:%.o=%.o.d)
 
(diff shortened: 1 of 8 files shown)


jserv (changes requested):
Instead of submitting a series of patches, let's work on decoupling x86-64 specific code out of the generic part first.

yanjiew1:
I have force pushed a new version that only contains the changes that separate x86-64 specific code out of generic implementation.

The original code is moved to another branch for backup.
https://github.com/yanjiew1/kvm-host/tree/arm64-final

src/arch/x86/vm-arch.h
```diff
@@ -0,0 +1,3 @@
+#pragma once
```
jserv: Rename this file to `src/arch/x86/desc.h` which stands for the architecture-specific descriptions.

src/arch/x86/vm.c
```diff
@@ -0,0 +1,209 @@
+#if !defined(__x86_64__) || !defined(__linux__)
+#error "This virtual machine requires Linux/x86_64."
```
jserv: Refine the error message as following:
> This implementation is dedicated to Linux/x86-64.

src/arch/x86/vm.c
```diff
@@ -0,0 +1,209 @@
...
+    boot->hdr.ramdisk_size = datasz;
+    return 0;
+}
+
+int vm_late_init(vm_t *v)
+{
+    return 0;
+}
+
+
```
jserv: Remove this blank line.

Makefile
```diff
@@ -21,6 +25,12 @@ OBJS := \
 	virtio-blk.o \
 	diskimg.o \
 	main.o
+
+ifeq ($(ARCH), x86_64)
+	CFLAGS += -I$(PWD)/src/arch/x86
+	OBJS += arch/x86/vm.o
```
jserv, line 32: Append
```c
CFLAGS += -include src/arch/x86/desc.h
```

src/vm.h
```diff
@@ -6,6 +6,7 @@
 #include "pci.h"
 #include "serial.h"
 #include "virtio-blk.h"
+#include "vm-arch.h"
```
jserv: Remove this inclusion since we don't want to expose the implementation details.

jserv:
Thank @yanjiew1 for contributing. Let's move forward to Arm64 specific commits.

PR #18, opened 2023-06-15
Enable fast string to boot on specific platform

RinHizakura:
I am not able to boot Linux completely with `kvm-host` on my new computer.

```
$ lscpu
Architecture:            x86_64
  CPU op-mode(s):        32-bit, 64-bit
  Address sizes:         39 bits physical, 48 bits virtual
  Byte Order:            Little Endian
CPU(s):                  16
  On-line CPU(s) list:   0-15
Vendor ID:               GenuineIntel
  Model name:            11th Gen Intel(R) Core(TM) i7-11700 @ 2.50GHz
```

After doing some research, I find that this could be a known issue on Intel's modern x86 CPUs with [Fast Short Rep Move](https://git.kernel.org/pub/scm/linux/kernel/git/tip/tip.git/commit/?h=x86/asm&id=f444a5ff95dce07cf4353cbb85fc3e785019d430) supported(FSRM=1).  In such a platform, if we don't enable fast string in `IA32_MISC_ENABLE`,  ERMS will be disabled in the boot process ([`early_init_intel`](https://elixir.bootlin.com/linux/latest/source/arch/x86/kernel/cpu/intel.c#L469)), which leads CPU feature become FSRM=1 and ERMS=0. However, the implementation of `memmove` is actually based on the assumption that ["FSRM implies ERMS"](https://elixir.bootlin.com/linux/latest/source/arch/x86/lib/memmove_64.S#L41), so we'll have trouble there. More detail and discussion of the issue can be found at [[PATCH] x86: also disable FSRM if ERMS is disabled]( https://lore.kernel.org/lkml/Y8J8yUfSKdLOjGW0@zn.tnic/T/). To fix this, we have to always enable the fast string.

By the way, since we don't support other platforms like ARM yet, is it fine to just put `vm_init_msrs` in `vm.c` now? Just remember to detach it elsewhere someday.

### DIFF
--- a/src/vm.c
+++ b/src/vm.c
@@ -73,6 +73,31 @@ static void vm_init_cpu_id(vm_t *v)
     ioctl(v->vcpu_fd, KVM_SET_CPUID2, &kvm_cpuid);
 }
 
+#define MSR_IA32_MISC_ENABLE 0x000001a0
+#define MSR_IA32_MISC_ENABLE_FAST_STRING_BIT 0
+#define MSR_IA32_MISC_ENABLE_FAST_STRING \
+    (1ULL << MSR_IA32_MISC_ENABLE_FAST_STRING_BIT)
+
+#define KVM_MSR_ENTRY(_index, _data)   \
+    (struct kvm_msr_entry)             \
+    {                                  \
+        .index = _index, .data = _data \
+    }
+static void vm_init_msrs(vm_t *v)
+{
+    int ndx = 0;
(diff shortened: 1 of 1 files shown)


jserv (changes requested):
Improve git commit message and make it informative.

jserv:
Thank @RinHizakura for contributing!

@yanjiew1, Be aware of the above changes while you are working on the separation of x86-64 specific code out of the generic implementation.

Issue #15, opened 2023-06-08
Refine UART emulation


jserv:
[tomu](https://github.com/thomasbarrett/tomu) emulates a 16550A UART device with `epoll`.
```
    epoll <──────── eventfd <─────╮
                                  │
               ╭──────────────╮   │
     read <─── │ │ │ fifo │ │ │ <─┴─ out
(nonblocking)  ╰──────────────╯
               ╭──────────────╮
    write ───> │ │ │ fifo │ │ │ ───> in
(nonblocking)  ╰──────────────╯
```

It would be great if we can learn from this project.

yanjiew1:
I have re-implemented the serial device using epoll and eventfd approach.
This is the pull request:
https://github.com/sysprog21/kvm-host/pull/16

PR #22, opened 2023-06-17
Refine FIFO operations of serial device

yanjiew1:
Instead of sending signal to notify the main thread to read more bytes into rx_buf, just do it in the worker thread.

When rx_buf is full and the worker thread cannot load more bytes, wait on the condition variable. The main thread will wake the worker thread when the level of the rx_buf decreases to its half size.

Since multiple threads can access the internal structure of the serial device, use a mutex to protect it.

It should be pushed to the original pull request #16, but it is closed when I renamed the branch. I did not find a way to reopen it, so I submit a new pull request.

### DIFF
--- a/src/serial.c
+++ b/src/serial.c
@@ -30,6 +30,8 @@ struct serial_dev_priv {
     uint8_t scr;
 
     struct fifo rx_buf;
+    pthread_mutex_t lock;
+    pthread_cond_t cond;
 };
 
 static struct serial_dev_priv serial_dev_priv = {
@@ -37,6 +39,8 @@ static struct serial_dev_priv serial_dev_priv = {
     .mcr = UART_MCR_OUT2,
     .lsr = UART_LSR_TEMT | UART_LSR_THRE,
     .msr = UART_MSR_DCD | UART_MSR_DSR | UART_MSR_CTS,
+    .lock = PTHREAD_MUTEX_INITIALIZER,
+    .cond = PTHREAD_COND_INITIALIZER,
 };
 
 /* FIXME: This implementation is incomplete */
@@ -52,7 +56,7 @@ static void serial_update_irq(serial_dev_t *s)
     else if ((priv->ier & UART_IER_THRI) && (priv->lsr & UART_LSR_TEMT))
         iir = UART_IIR_THRI;
 
-    priv->iir = iir | 0xc0;
+    __atomic_store_n(&priv->iir, iir | 0xc0, __ATOMIC_RELEASE);
 
     /* FIXME: the return error of vm_irq_line should be handled */
     vm_irq_line(container_of(s, vm_t, serial), s->irq_num,
@@ -76,9 +80,19 @@ static volatile bool thread_stop = false;
 
 static void *serial_thread(serial_dev_t *s)
 {
+    struct serial_dev_priv *priv = (struct serial_dev_priv *) s->priv;
     while (!__atomic_load_n(&thread_stop, __ATOMIC_RELAXED)) {
-        if (serial_readable(s, -1))
-            pthread_kill((pthread_t) s->main_tid, SIGUSR1);
+        if (!serial_readable(s, -1))
+            continue;
+        pthread_mutex_lock(&priv->lock);
+        if (fifo_is_full(&priv->rx_buf)) {
+            /* stdin is readable, but the rx_buf is full.
+             * Wait for notification.
+             */
+            pthread_cond_wait(&priv->cond, &priv->lock);
+        }
+        serial_console(s);
+        pthread_mutex_unlock(&priv->lock);
     }
 
     return NULL;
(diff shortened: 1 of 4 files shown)


src/serial.c
```diff
@@ -76,9 +80,17 @@ static volatile bool thread_stop = false;
 
 static void *serial_thread(serial_dev_t *s)
 {
+    struct serial_dev_priv *priv = (struct serial_dev_priv *) s;
     while (!__atomic_load_n(&thread_stop, __ATOMIC_RELAXED)) {
-        if (serial_readable(s, -1))
-            pthread_kill((pthread_t) s->main_tid, SIGUSR1);
+        if (serial_readable(s, -1)) {
```
jserv: Rewrite as following:
```c
    if (!serial_readable(s, -1))
        continue;

    pthread_mutex_lock(&priv->lock);
    ...
```

src/serial.c
```diff
@@ -122,6 +131,10 @@ static void serial_in(serial_dev_t *s, uint16_t offset, void *data)
                 priv->lsr &= ~UART_LSR_DR;
                 serial_update_irq(s);
             }
+            if (fifo_level(&priv->rx_buf) == FIFO_LEN / 2)
+                /* rx_buf is half full now. Notify the worker thread to read
```
jserv: You shall explain the reason why half full matters.

src/serial.c
```diff
@@ -204,29 +217,17 @@ static void serial_handle_io(void *owner,
                              uint8_t size)
 {
     serial_dev_t *s = (serial_dev_t *) owner;
+    struct serial_dev_priv *priv = (struct serial_dev_priv *) s;
     void (*serial_op)(serial_dev_t *, uint16_t, void *) =
         is_write ? serial_out : serial_in;
 
+    pthread_mutex_lock(&priv->lock);
```
RinHizakura: It looks like this critical section only races with `serial_console`, which only access `rx_buf` there. Can we just lock smaller sections to improve the concurrency?

Makefile
```diff
@@ -3,12 +3,14 @@ include mk/common.mk
 ARCH ?= $(shell uname -m)
 PWD ?= $(shell pwd)
 
+SANITIZE = -fsanitize=thread
```
jserv: Don't do that. Minimize the necessary changes.

src/serial.c
```diff
@@ -76,9 +80,17 @@ static volatile bool thread_stop = false;
...
 static void *serial_thread(serial_dev_t *s)
 {
+    struct serial_dev_priv *priv = (struct serial_dev_priv *) s->priv;
     while (!__atomic_load_n(&thread_stop, __ATOMIC_RELAXED)) {
-        if (serial_readable(s, -1))
-            pthread_kill((pthread_t) s->main_tid, SIGUSR1);
+        if (!serial_readable(s, -1))
+            continue;
+        pthread_mutex_lock(&priv->lock);
+        if (fifo_is_full(&priv->rx_buf))
```
jserv: Rewrite as following:
```c
        if (fifo_is_full(&priv->rx_buf)) {
            /* stdin is readable, but the rx_buf is full.
             * Wait for notification.
             */
            pthread_cond_wait(&priv->cond, &priv->lock);
        }
```

Clarify for multiline comment and statement.

src/utils.h
```diff
@@ -19,6 +19,8 @@ struct fifo {
 
 #define fifo_is_full(fifo) ((fifo)->tail - (fifo)->head > FIFO_MASK)
 
+#define fifo_level(fifo) ((fifo)->tail - (fifo)->head)
```
jserv: The naming is a bit misleading. Did you mean `fifo_capacity`?

src/serial.c
```diff
@@ -88,40 +100,44 @@ void serial_console(serial_dev_t *s)
...
             if (fifo_get(&priv->rx_buf, value))
                 IO_WRITE8(data, value);
 
             if (fifo_is_empty(&priv->rx_buf)) {
                 priv->lsr &= ~UART_LSR_DR;
                 serial_update_irq(s);
             }
+            /* The worker thread waits on the condition variable when rx_buf is
+             * full and stdin is still readable. Notify the worker thread when
+             * the level of the buffer drops to half of its size, so the worker
```
jserv: Ditto. Refine the term "level" properly.

src/serial.c
yanjiew1: I've added more explanation to this comment. This is the new revision taken from the updated source code:
```
/* The worker thread waits on the condition variable when rx_buf is
 * full and stdin is still readable. Notify the worker thread when
 * the level of the buffer drops to its half size, so the worker
 * thread can read up to half of the buffer size before it is
 * blocked again.
 */
```
Could you please review it?

src/serial.c
yanjiew1: I've made the critical section smaller.
Because not only `rx_buf`, but `iir`, `ier` and `lsr` can be accessed by the worker thread and the main thread, some loads and stores to `iir` and `lsr` are modified to use atomic operations.

Makefile
yanjiew1: I added it accidentally. I've removed this change in the new revision.

src/serial.c
jserv: It looks better now. Refine the term "level" as I requested.

jserv (changes requested):
Append `Close #15` at the end of git commit message since the proposed changes would resolve.

jserv:

PR #26, opened 2023-06-19
Add initial support for arm64

yanjiew1:
This pull request adds initial support for arm64.

Since each platform may define different IRQ numbers and kernel cmdline, I moved the definitions of IRQ numbers and kernel cmdline to architecture-specific `desc.h`.

### DIFF
--- a/src/arch/arm64/vm-arch.h
+++ b/src/arch/arm64/vm-arch.h
@@ -1,0 +1,51 @@
+#pragma once
+
+#ifndef SZ_64K
+#define SZ_64K (1UL << 16)
+#endif
+
+#include <asm/kvm.h>
+
+/*
+ * The maximum size of the device tree is 2MB.
+ * Reference: https://docs.kernel.org/arm64/booting.html
+ */
+#define FDT_MAX_SIZE (1UL << 21)
+
+/*
+ *  Memory map for guest memory
+ *
+ *    0 -  64K  I/O Ports
+ *   1M -  16M  GIC
+ *  1GB -  2GB  PCI MMIO
+ *  2GB -       DRAM
+ */
+
+#define ARM_IOPORT_BASE 0
+#define ARM_IOPORT_SIZE (1UL << 16)
+
+#define ARM_GIC_BASE 0x100000UL
+
+#define ARM_GIC_DIST_BASE ARM_GIC_BASE
+#define ARM_GIC_DIST_SIZE KVM_VGIC_V3_DIST_SIZE
+
+#define ARM_GIC_REDIST_BASE (ARM_GIC_DIST_BASE + ARM_GIC_DIST_SIZE)
+#define ARM_GIC_REDIST_SIZE KVM_VGIC_V3_REDIST_SIZE
+
+#define ARM_PCI_CFG_BASE 0x40000000UL
+#define ARM_PCI_CFG_SIZE (1UL << 16)
+
+#define ARM_PCI_MMIO_BASE (ARM_PCI_CFG_BASE + ARM_PCI_CFG_SIZE)
+#define ARM_PCI_MMIO_SIZE (RAM_BASE - ARM_PCI_MMIO_BASE)
+
+/* 128 MB for iernel */
+#define ARM_KERNEL_BASE RAM_BASE
+#define ARM_KERNEL_SIZE 0x8000000UL
+
+/* 128 MB for initrd */
+#define ARM_INITRD_BASE (ARM_KERNEL_BASE + ARM_KERNEL_SIZE)
+#define ARM_INITRD_SIZE 0x8000000UL
+
+/* For FTB */
+#define ARM_FDT_BASE (ARM_INITRD_BASE + ARM_INITRD_SIZE)
+#define ARM_FDT_SIZE FDT_MAX_SIZE
--- a/src/arch/arm64/vm.c
+++ b/src/arch/arm64/vm.c
@@ -1,0 +1,463 @@
+#if !defined(__aarch64__) || !defined(__linux__)
+#error "This implementation is dedicated to Linux/aarch64."
+#endif
+
+#include <libfdt.h>
+#include <linux/kvm.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <string.h>
+#include <sys/ioctl.h>
+
+#include "err.h"
+#include "vm-arch.h"
+#include "vm.h"
+
+typedef struct {
+    uint64_t entry;
+    size_t initrdsz;
+    int gic_fd;
+
+    /* This device is a bridge between mmio_bus and io_bus*/
+    struct dev iodev;
+} vm_arch_priv_t;
+
+static vm_arch_priv_t vm_arch_priv;
+
+static int create_irqchip(vm_t *v)
+{
+    vm_arch_priv_t *priv = (vm_arch_priv_t *) v->priv;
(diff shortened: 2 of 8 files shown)


src/arch/arm64/vm.c
```diff
@@ -0,0 +1,490 @@
...
+struct vm_arch_priv {
+    uint64_t entry;
+    size_t initrdsz;
+    int gic_fd;
+
+    /* This device is a bridge between mmio_bus and io_bus*/
+    struct dev iodev;
+};
+
+struct vm_arch_priv vm_arch_priv;
```
jserv: Declare as `static`.

src/arch/arm64/vm.c
```diff
@@ -0,0 +1,490 @@
...
+#include <sys/ioctl.h>
+#include <sys/mman.h>
+#include <sys/stat.h>
+#include <unistd.h>
+
+#include "err.h"
+#include "vm-arch.h"
+#include "vm.h"
+
+struct vm_arch_priv {
```
jserv: Make use of typedef. i.e.,
```c
typedef struct {
    uint64_t entry;
    size_t initrdsz;
    int gic_fd;

    /* This device is a bridge between mmio_bus and io_bus*/
    struct dev iodev;
} vm_arch_priv_t;
```

Then, replace all occurrence `strut vm_arch_priv` with `vm_arch_priv_t`.

src/arch/arm64/vm.c
```diff
@@ -0,0 +1,490 @@
...
+#include <fcntl.h>
+#include <libfdt.h>
+#include <linux/kvm.h>
+#include <linux/kvm_para.h>
+#include <stdbool.h>
+#include <stddef.h>
+#include <stdint.h>
+#include <string.h>
+#include <sys/ioctl.h>
+#include <sys/mman.h>
```
jserv: Did you invoke `mmap`? If not, clean up the header inclusions.

yanjiew1:
I pushed a new revision.
In addition to the above changes, I also removed the "/psci" node in the device tree since it is not required for Linux to boot.

src/arch/arm64/vm.c
```diff
@@ -0,0 +1,469 @@
...
+#include <stddef.h>
+#include <stdint.h>
+#include <string.h>
+#include <sys/ioctl.h>
+
+#include "err.h"
+#include "vm-arch.h"
+#include "vm.h"
+
+typedef struct vm_arch_priv {
```
jserv: Replace `typedef struct vm_arch_priv` with `typedef struct`. (anonymous structure)

src/arch/arm64/vm.c
```diff
@@ -0,0 +1,469 @@
...
+        return throw_err("Failed to initialize vCPU\n");
+
+    return 0;
+}
+
+/* This should be called after all vCPUs are created */
+static int finalize_irqchip(vm_t *v)
+{
+    vm_arch_priv_t *priv = (vm_arch_priv_t *) v->priv;
+    int nirqs = 992;
```
jserv: Can you explain why 992 appears here?

yanjiew1: It is the maximum number of IRQs the guest supported, including SGIs and PPIs.

The document in kernel source says that the maximum allowed value can be 1024.
I tried 1024 first, but I found that it cannot work. After some trial and error, I found that the maximum allowed value is 992..
https://www.kernel.org/doc/Documentation/virtual/kvm/devices/arm-vgic-v3.txt

I think that we could remove this setting as the default is 256 according to the kernel source, which should be enough.

src/arch/arm64/vm-arch.h
```diff
@@ -0,0 +1,49 @@
+#pragma once
+
+#define SZ_64K (1UL << 16)
```
jserv, line 4: Is this macro used?

jserv: If this macro is required, you shall wrap it before using. i.e.,
```c
#ifdef SZ_64K
#define SZ_64K (1UL << 16)
#endif
```

yanjiew1: This macro is used by other macros in the header file "asm/kvm.h".
I pushed a new revision and wrap `#define SZ_64K ...` as following:
```c
#ifndef SZ_64K
#define SZ_64K (1UL << 16)
#endif
```

You mentioned `#ifdef`, which I think is a typo.

src/arch/arm64/vm.c
```diff
@@ -0,0 +1,459 @@
...
+
+    if (ioctl(v->vcpu_fd, KVM_GET_ONE_REG, &reg) < 0)
+        return throw_err("Failed to get MPIDR register\n");
+
+    *mpidr &= ARM_MPIDR_BITMASK;
+    return 0;
+}
+
+/* Helper function to simplify error handling */
+#define _FDT(exp)                                                         \
```
jserv: Furthermore, you can utilize [Variadic Macros](https://gcc.gnu.org/onlinedocs/cpp/Variadic-Macros.html) to make it more elegant.

Original
```c
_FDT(fdt_create(fdt, FDT_MAX_SIZE));
```

Revised:
```c
__FDT(create, FDT_MAX_SIZE);
```

The possible definition of the above macro would be:
```c
#define __FDT(action, ...) \
    do {
        int __ret = fdt_##action(fdt, __VA_ARGS__ );
        if (__ret < 0)
            return throw_err(...);
    } while (0)
```

By the way, the definition of `__FDT` should appear right before the function definition of `generate_fdt`, and there must be a corresponding `#undef __FDT` right after `generate_fdt`.

jserv:
Once this pull request is merged, please work on the generation of Linux kernel and busybox based on the supported platforms. That is, you shall provide the following:
1. Aarch64 specific kernel configurations;
2. top-level documentation which addresses the supported architectures;
3. the consistent way to run `make check` on both x86-64 and Aarch64.

src/arch/arm64/vm.c
jserv: Alternatively, the way what you have used for `_REG` is reasonably well. You can  apply `__FDT` in the same way. For consistency, enforce `__REG` rather than `_REG`.

yanjiew1: I pushed a new revision with these changes.

src/arch/arm64/vm.c
```diff
@@ -0,0 +1,462 @@
...
+#define ARM_FDT_IRQ_LEVEL_TRIGGER 4
+
+/* Helper macro to simplify error handling */
+#define __FDT(action, ...)                                                  \
+    do {                                                                    \
+        int __ret = fdt_##action(fdt, ##__VA_ARGS__);                       \
+        if (__ret >= 0)                                                     \
+            break;                                                          \
+        return throw_err("Failed to create device tree:\n %s\n %s\n",       \
+                         "_FDT(" #action __VA_OPT__(", ") #__VA_ARGS__ ")", \
```
jserv: The message "_FDT(" #action __VA_OPT__(" is boring.
You can simply alert something like `fdt_property_cell` without mentioning `_FDT("

yanjiew1: Should I include the parameters passed to the fdt_... function in the error message, like `fdt_property_cell(fdt, .....)`? Or just simply mention the function name? `fdt_property_cell`

yanjiew1: I pushed a new revision. The error message contains both function name and the parameters.

jserv:
Thank @yanjiew1 for this effort!

PR #27, opened 2023-06-20
Add arm64 support for "make check"

yanjiew1:
This pull request add arm64 support for "make check".

I added a new kernel config file for arm64.
The command used to configure the kernel is changed from "make oldconfig" to "make olddefconfig" as "make oldconfig" may ask users to confirm configs that are not included in ".config" file.

The arm64 kernel configuration file is generated by using the command, "make savedefconfig", which can be used to generate a minimal configuration file with the filename, "defconfig". This file can be renamed to ".config", and then we can use "make olddefconfig" to convert it to a full configuration file.


mk/external.mk
```diff
@@ -34,14 +34,23 @@ EXTERNAL_SRC = LINUX BUSYBOX
 $(foreach T,$(EXTERNAL_SRC),$(eval $(download-n-extract)))
 
 # Build Linux kernel image
-LINUX_IMG = $(OUT)/bzImage
+ifeq ($(ARCH), x86_64)
+LINUX_IMG_NAME = bzImage
+LINUX_ARCH = x86
+endif
+ifeq ($(ARCH), aarch64)
```
jserv: Replace this line with `else ifeq ($(ARCH), aarch64)`.
In addition, append the following:
```
else
    $(error Unsupported architecture)
endif
```

jserv (changes requested):
Use the naming scheme `configs/linux-aarch64.config` and `configs/linux-x86_64.config`

jserv:
By using `olddefconfig`, we can shorten the configures for x86-64. Let's do that in further pull request(s).

mk/external.mk
```diff
@@ -34,14 +34,24 @@ EXTERNAL_SRC = LINUX BUSYBOX
...
+ifeq ($(ARCH), x86_64)
+LINUX_IMG_NAME = bzImage
+LINUX_ARCH = x86
+else ifeq ($(ARCH), aarch64)
+LINUX_IMG_NAME = Image
+LINUX_ARCH = arm64
+else
+    $(error Unsupported architecture)
+endif
+LINUX_IMG = $(OUT)/$(LINUX_IMG_NAME)
```
jserv: Rewrite as following:
```
LINUX_IMG := $(addprefix $(OUT)/,$(LINUX_IMG))
```

It is a matter of variable overriding. See https://www.gnu.org/software/make/manual/html_node/Overriding.html

mk/external.mk
```diff
@@ -34,14 +34,24 @@ EXTERNAL_SRC = LINUX BUSYBOX
 $(foreach T,$(EXTERNAL_SRC),$(eval $(download-n-extract)))
 
 # Build Linux kernel image
-LINUX_IMG = $(OUT)/bzImage
+ifeq ($(ARCH), x86_64)
+LINUX_IMG_NAME = bzImage
+LINUX_ARCH = x86
```
jserv: Maybe we can modify `$(ARCH)` directly, so that we don't have to deal with `LINUX_ARCH`.

yanjiew1: I changed `LINUX_ARCH` to `ARCH` which means that x86 or arm64 will be assigned to `ARCH`.

However, if we directly modify `$(ARCH)`, the filename of Linux kernel configuration file should become `linux-x86.config` and `linux-arm64.config` as the variable `$(ARCH)` is used in the filename.

mk/external.mk
yanjiew1: I pushed a new revision and changed `LINUX_IMG = $(OUT)/$(LINUX_IMG_NAME)` to `LINUX_IMG := $(addprefix $(OUT)/,$(LINUX_IMG_NAME))`

I kept the variable `LINUX_IMG_NAME` because it is used when compiling the kernel and copying the output file.

Issue #25, opened 2023-06-19
Handle Ctrl-x a sequence to exit


jserv:
Currently, when Ctrl-C is pressed, kvm-host terminates unexpectedly, leading to usability issues. To address this problem, we can implement the same behavior as QEMU, where the key sequence Ctrl-X followed by 'A' is used.

Reference:
* [Handle UART input with termios](https://github.com/jserv/semu/pull/17)

yanjiew1:
The key sequence to terminate QEMU should be Ctrl-a x.

I have sent a pull request #30, which implements the key sequence ctrl-a x.

PR #30, opened 2023-06-28
Handle Ctrl-a x sequence to exit

yanjiew1:
Originally, the guest will be terminated when ctrl-c is pressed. This causes the guest to exit unexpectedly.

This change addresses this issue by removing ISIG flag in c_lflag in termios and adds a key sequence ctrl-a x to terminate the guest.

### DIFF
--- a/src/serial.c
+++ b/src/serial.c
@@ -97,14 +97,28 @@ static void *serial_thread(serial_dev_t *s)
     return NULL;
 }
 
+#define TERMINAL_ESCAPE_CHAR 0x01
+#define TERMINAL_EXIT_CHAR 'x'
+
 void serial_console(serial_dev_t *s)
 {
     struct serial_dev_priv *priv = (struct serial_dev_priv *) s->priv;
+    static bool escaped = false;
 
     while (!fifo_is_full(&priv->rx_buf) && serial_readable(s, 0)) {
         char c;
         if (read(s->infd, &c, 1) == -1)
             break;
+        if (escaped && c == TERMINAL_EXIT_CHAR) {
+            /* Terminate */
+            fprintf(stderr, "\n");
+            exit(0);
+        }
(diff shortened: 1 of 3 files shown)


src/serial.c
```diff
@@ -97,14 +97,28 @@ static void *serial_thread(serial_dev_t *s)
     return NULL;
 }
 
+#define TERMINAL_ESCAPE_CHAR 0x01
+#define TERMINAL_TERMINATE_CHAR 'x'
```
jserv: Shorten it as `TERMINAL_EXIT_CHAR`.

src/serial.c
```diff
@@ -97,14 +97,28 @@ static void *serial_thread(serial_dev_t *s)
...
     struct serial_dev_priv *priv = (struct serial_dev_priv *) s->priv;
+    static bool escaped = false;
 
     while (!fifo_is_full(&priv->rx_buf) && serial_readable(s, 0)) {
         char c;
         if (read(s->infd, &c, 1) == -1)
             break;
+        if (escaped && c == TERMINAL_TERMINATE_CHAR) {
+            /* Terminate */
+            fprintf(stderr, "\nTerminated\n");
```
jserv: I am not if it makes sense to dump this messages.

jserv (changes requested):
Add top-level documentation about the proposed key sequence in `README.md`.

jserv (changes requested):
Append "Close #25" at the end of git commit message.

src/serial.c
yanjiew1: I changed it to `fprintf(stderr, "\n");`.
The newline character is inserted just before kvm-host is terminated.  This ensures that the shell prompt starts at the beginning of a line.

jserv:

PR #31, opened 2023-10-05
Add git submodule in build section in README

howjmay:
README didn't mention git submodules are required when building this
project.


jserv:
The submodule only makes sense to Arm build.

howjmay:
> The submodule only makes sense to Arm build.

I have added the description that it is only necessary for ARM build.

PR #34, opened 2024-06-27
Implement virtio-net device

jimmylu890303:
## Implementation of Virtio-net Device
- Defined struct `virtio_net_dev` to represent the virtio-net device.
- Implemented `virtio_net_init` to initialize the TAP device for communication.
- Introduced `virtio_net_init_pci` to register the virtio-net device on PCI, configure virtqueues.
- Implemented functions to handle available events for RX and TX virtqueues.
- Added functions to notify the guest OS about used descriptors after packet processing.

## Makefile Modification
- Added rules to compile virtio-net.c into virtio-net.o, ensuring correct compilation and linkage of virtio-net functionality within the project.
## Kernel Networking Options Enabled
-  Enabled necessary kernel configuration options (`CONFIG_NET`, `CONFIG_INET`, `CONFIG_VIRTIO`, `CONFIG_VIRTIO_NET`, `CONFIG_NETDEVICES`) to support networking functionality.
- Enabled specific networking commands required for testing the virtio-net device.

## Structural Updates
- Added struct `virtio_net_dev` entry within struct `vm_t` to facilitate the integration and management of the virtio-net device.

## Additional Changes
- Called `virtio_net_init` and `virtio_net_init_pci` in `vm_arch_init_platform_device` and `main` respectively, ensuring initialization of the virtio-net device before the virtual machine starts.


Makefile
```diff
@@ -31,7 +31,8 @@ OBJS := \
 	virtq.o \
 	virtio-blk.o \
 	diskimg.o \
-	main.o
+	main.o \
+	virtio-net.o
```
jserv: Move `virtio-net.o` right after the line containing `virtio-blk.o`.

configs/busybox.config
```diff
@@ -847,16 +847,16 @@ CONFIG_TREE=y
 #
 # Networking Utilities
 #
-# CONFIG_FEATURE_IPV6 is not set
+CONFIG_FEATURE_IPV6=y
```
jserv: Disable IPv6 for simplicity.

configs/busybox.config
```diff
@@ -888,21 +888,21 @@ CONFIG_FEATURE_HTTPD_PORT_DEFAULT=0
 # CONFIG_FEATURE_HTTPD_LAST_MODIFIED is not set
 # CONFIG_FEATURE_HTTPD_DATE is not set
 # CONFIG_FEATURE_HTTPD_ACL_IP is not set
-# CONFIG_IFCONFIG is not set
-# CONFIG_FEATURE_IFCONFIG_STATUS is not set
-# CONFIG_FEATURE_IFCONFIG_SLIP is not set
-# CONFIG_FEATURE_IFCONFIG_MEMSTART_IOADDR_IRQ is not set
-# CONFIG_FEATURE_IFCONFIG_HW is not set
+CONFIG_IFCONFIG=y
```
jserv, line 891: `ifconfig` was deprecated. Always use `ip` command.

configs/linux-x86.config
```diff
@@ -115,3 +115,24 @@ CONFIG_FRAME_WARN=1024
 # CONFIG_X86_DEBUG_FPU is not set
 CONFIG_UNWINDER_GUESS=y
 # CONFIG_RUNTIME_TESTING_MENU is not set
+
```
jserv: Drop these blank lines.

configs/linux-x86.config
```diff
@@ -115,3 +115,24 @@ CONFIG_FRAME_WARN=1024
...
+
+# Networking options
```
jserv: The comment is not necessary.

configs/linux-x86.config
```diff
@@ -115,3 +115,24 @@ CONFIG_FRAME_WARN=1024
...
+CONFIG_NET=y
+CONFIG_INET=y
+# CONFIG_IPV6 is not set  
+# CONFIG_IPV6_SIT is not set  
+CONFIG_IP_MULTICAST=y
+CONFIG_IP_PNP=y
+CONFIG_IP_PNP_DHCP=y
+
```
jserv: Drop blank lines in this configuration file.

src/arch/x86/desc.h
```diff
@@ -3,4 +3,5 @@
 #define RAM_BASE 0
 #define SERIAL_IRQ 4
 #define VIRTIO_BLK_IRQ 15
+#define VIRTIO_NET_IRQ 14
```
jserv, line 5: Swap the order between `VIRTIO_BLK_IRQ` and `VIRTIO_NET_IRQ`.

src/arch/x86/vm.c
```diff
@@ -96,7 +96,7 @@ int vm_arch_init(vm_t *v)
     __u64 map_addr = 0xffffc000;
     if (ioctl(v->vm_fd, KVM_SET_IDENTITY_MAP_ADDR, &map_addr) < 0)
         return throw_err("Failed to set identity map address");
-
+
```
jserv: Don't change this. Minimize the necessary changes.

src/virtio-net.c
```diff
@@ -0,0 +1,292 @@
...
+    struct pollfd pollfd = (struct pollfd){
+        .fd = dev->tapfd,
+        .events = POLLIN,
+    };
+    return (poll(&pollfd, 1, timeout) > 0) && (pollfd.revents & POLLIN);
+}
+
+static int virtio_net_virtq_available_tx(struct virtio_net_dev *dev, int timeout)
+{
+    struct pollfd pollfds[2];
```
jserv: Rewrite using C99 designated initializers. That is,
```c
    struct pollfd pollfds[] = {
        [0] = {.fd = dev->tx_ioeventfd, events = POLLIN},
        [1] = {.fd = dev->tapfd, .events = POLLOUT},
    };
```

src/virtio-net.c
```diff
@@ -0,0 +1,292 @@
...
+    pollfds[1].fd = dev->tapfd;
+    pollfds[1].events = POLLOUT;
+
+    int ret = poll(pollfds, 2, timeout);
+
+    return ret > 0 && (pollfds[0].revents & POLLIN) && (pollfds[1].revents & POLLOUT);
+}
+
+void sigusr1_handler(int signum) {
+    return ;
```
jserv: Remove `return` statement here. Mark `sigusr1_handler` as a static function.

src/virtio-net.c
```diff
@@ -0,0 +1,292 @@
...
+}
+
+static void *virtio_net_vq_avail_handler_rx(void *arg){
+    struct virtq *vq = (struct virtq *)arg;
+    struct virtio_net_dev *dev = (struct virtio_net_dev *)vq->dev;
+
+    struct sigaction sa;
+    sa.sa_handler = sigusr1_handler;
```
jserv: You should explain why you would like to hook SIGUSR1.

src/virtio-net.c
```diff
@@ -0,0 +1,292 @@
...
+    strncpy(ifreq.ifr_name, TAP_INTERFACE, sizeof(ifreq.ifr_name));
+    if (ioctl(virtio_net_dev->tapfd, TUNSETIFF, &ifreq) < 0) {
+        fprintf(stderr, "failed to allocate TAP device: %s\n", strerror(errno));
+        return false;
+    }
+    assert(fcntl(virtio_net_dev->tapfd, F_SETFL,
+                fcntl(virtio_net_dev->tapfd, F_GETFL, 0) | O_NONBLOCK) >= 0);
+    return true;
+}
+static void virtio_net_setup(struct virtio_net_dev *dev)
```
jserv, line 237: Add a blank line fore the implementation of `virtio_net_setup`.

src/virtio-net.h
```diff
@@ -0,0 +1,32 @@
...
+    bool enable;
+};
+
+bool virtio_net_init(struct virtio_net_dev *virtio_net_dev);
+void virtio_net_exit(struct virtio_net_dev *virtio_net_dev);
+void virtio_net_init_pci(struct virtio_net_dev *virtio_net_dev,
+                            struct pci *pci,
+                            struct bus *io_bus,
+                            struct bus *mmio_bus);
+
```
jserv: Remove the blank line at the end of this file.

src/virtio-pci.h
```diff
@@ -7,6 +7,7 @@
 
 #define VIRTIO_PCI_VENDOR_ID 0x1AF4
 #define VIRTIO_PCI_DEVICE_ID_BLK 0x1042
+#define VIRTIO_PCI_DEVICE_ID_NET 0x1041
```
jserv, line 9: Swap the order between `VIRTIO_PCI_DEVICE_ID_NET` and `VIRTIO_PCI_DEVICE_ID_BLK`.

src/main.c
```diff
@@ -93,7 +93,7 @@ int main(int argc, char *argv[])
         return throw_err("Failed to load initrd");
     if (diskimg_file && vm_load_diskimg(&vm, diskimg_file) < 0)
         return throw_err("Failed to load disk image");
-
+    virtio_net_init_pci(&vm.virtio_net_dev, &vm.pci, &vm.io_bus, &vm.mmio_bus);
```
RinHizakura: We should provide a consistent interface between virtio_blk and virtio_net, which lets us manage them better when more I/O devices come in the future.

Add an API `int vm_net_init(vm_t *v)` and use it here. Place `virtio_net_init_pci()` in `vm_enable_net()`.

src/virtio-net.c
```diff
@@ -0,0 +1,292 @@
...
+    vm_t *v = container_of(dev, vm_t, virtio_net_dev);
+
+    dev->enable = true;
+    dev->irq_num = VIRTIO_NET_IRQ;
+    dev->rx_ioeventfd = eventfd(0, EFD_CLOEXEC);
+    dev->tx_ioeventfd = eventfd(0, EFD_CLOEXEC);
+    dev->irqfd = eventfd(0, EFD_CLOEXEC);
+    vm_irqfd_register(v, dev->irqfd, dev->irq_num, 0);
+    for (int i = 0; i < VIRTIO_NET_VIRTQ_NUM; i++){
+        struct virtq_ops *ops = (i == 0) ? &virtio_net_rx_ops : &virtio_net_tx_ops;
```
RinHizakura: I'll suggest to declare an array for `rx_ops` and `tx_ops` and iterate here.

```cpp
#define VIRTQ_RX 0
#define VIRTQ_TX 1

static struct virtq_ops virtio_net_ops[VIRTIO_NET_VIRTQ_NUM] = {
    [VIRTQ_RX] = {
        .enable_vq = virtio_net_enable_vq_rx,
       .complete_request = virtio_net_complete_request_rx,·
       .notify_used = virtio_net_notify_used_rx,
    },
    [VIRTQ_TX] = {
        .enable_vq = virtio_net_enable_vq_tx,
       .complete_request = virtio_net_complete_request_tx,·
       .notify_used = virtio_net_notify_used_tx,
    },
}

...

    for (int i = 0; i < VIRTIO_NET_VIRTQ_NUM; i++){
        struct virtq_ops *ops = &virtio_net_ops[i];
```

src/virtio-net.c
```diff
@@ -0,0 +1,292 @@
...
+    vq->desc_ring = 
+        (struct vring_packed_desc *) vm_guest_to_host(v, vq->info.desc_addr);
+    vq->device_event = (struct vring_packed_desc_event *) vm_guest_to_host(
+            v, vq->info.device_addr);
+    vq->guest_event = (struct vring_packed_desc_event *) vm_guest_to_host(
+            v, vq->info.driver_addr);
+ 
+    uint64_t addr = virtio_pci_get_notify_addr(&dev->virtio_pci_dev, vq);
+    vm_ioeventfd_register(v, dev->tx_ioeventfd, addr, 
+                          2, 0);
```
RinHizakura: Why do we use a 2 here instead of  `sizeof(struct virtio_pci_notify_data)`/`notify_cap->cap.length`? Also please don't use magic numbers if possible.

jimmylu890303: Because `cap.length` is 4, the lower 2 bytes are used to notify the device from the `rx_virtqueue`, while the upper 2 bytes are used to notify the device from the `tx_virtqueue`.

Therefore, when calling `virtio_pci_get_notify_addr`, the address of the `rx_virtqueue` is `cap.offset` (assuming `notify_off_multiplier=2` and `vq->info.notify_off=0`), and the address of the `tx_virtqueue` is `notify_off_multiplier * notify_off` (assuming `notify_off_multiplier=2` and `vq->info.notify_off=1`). This allows me to register these two addresses for `rx_ioeventfd` and `tx_ioeventfd`.

src/virtio-net.c
jimmylu890303: Yes, I have already made the changes as you suggested.

src/main.c
jimmylu890303: I have added `vm_net_init`, which will be called in `vm_arch_init_platform_device` to initialize the tap device. Additionally, `vm_enable_net` will be called in the initial session of the VM to register a virtio-net device on PCI.

src/virtio-pci.h
jimmylu890303: Yes, I have already swapped the order.

src/virtio-net.h
jimmylu890303: I have already removed the blank.

src/virtio-net.c
jimmylu890303: I have already added the blank before `virtio_net_setup`.

src/virtio-net.c
jimmylu890303: I have already removed `sigusr1_handler` because there is no need to hook SIGUSR1. It was unnecessary in the previous version.

src/virtio-net.c
jimmylu890303: I have already changed to the C99 style.

src/arch/x86/desc.h
jimmylu890303: I have already swapped the order.

Makefile
jimmylu890303: I have already changed the order.

configs/busybox.config
jimmylu890303: I have disabled it.

configs/linux-x86.config
jimmylu890303: I have already dropped

configs/linux-x86.config
jimmylu890303: I have already removed

configs/linux-x86.config
jimmylu890303: I have already dropped

jimmylu890303:
I have pushed a newer version. Please take a look.

For testing the function, you can use the command below:

Host OS :
```
sudo ip link delete br0
sudo brctl addbr br0
sudo ip addr add 10.0.0.1/24 dev br0
sudo ip route add default via 10.0.0.1 dev br0
sudo ip link set br0 up
sudo ip link set tap0 master br0
sudo ip link set tap0 up
```
Guest OS : 
```
ip addr add 10.0.0.2/24 dev eth0
ip link set eth0 up
ip route add default via 10.0.0.1
```

src/main.c
RinHizakura: I mean add an API `int vm_net_init(vm_t *v)` and use it here, and placing `virtio_net_init_pci()` in `vm_net_vinit()` will be good enough. Sorry for the confusing comment. :(

src/virtio-net.c
RinHizakura: AFAIK, our `virtio-pci` defines a general notify data which is mapped to the `struct virtio_pci_notify_data`.
```
struct virtio_pci_notify_data {
    uint16_t vqn;
    uint16_t next;
} __attribute__((packed));
```
Does it mean virtio-net uses the first 2 bytes for the rx event and the last 2 bytes for the tx event on this structure? Sorry if I misunderstand this.

jimmylu890303: If we don't set the `notify_off_multiplier`, when `virtio_net_enable_vq_rx` and `virtio_net_enable_vq_tx` call `uint64_t addr = virtio_pci_get_notify_addr(&dev->virtio_pci_dev, vq)`, it causes the RX and TX to get the same notify address. This results in a problem where the ioeventfd is registered at the same address repeatedly.

To fix this, I changed the `notify_off_multiplier` to 2. Now, `rx_virtq.notify_off` is 0 and `tx_virtq.notify_off` is 1, allowing me to register 2 eventfds at these 2 distinct addresses.

RinHizakura: According to my understanding, all the virtio devices registered through `virtio_pci_init` have a general MMIO space layout(the `struct virtio_pci_config`) and a general notify data(the `struct virtio_pci_notify_data`). As you can see, we map the first 2 bytes of notify data to `vqn` and the last 2 bytes to `next`. But in your case it looks like you are mapping the first 2 bytes of notify data to `rx_vqn` and the last 2 bytes to `tx_vqn`. Am I correct?

src/arch/x86/vm.c
```diff
@@ -123,6 +129,15 @@ int vm_arch_init_platform_device(vm_t *v)
     if (serial_init(&v->serial, &v->io_bus))
         return throw_err("Failed to init UART device");
     virtio_blk_init(&v->virtio_blk_dev);
+    vm_net_init(&v->virtio_net_dev);
```
RinHizakura: It should be fine to `virtio_net_init()` here directly. Remove `vm_net_init()`.

src/arch/x86/vm.c
```diff
@@ -123,6 +129,15 @@ int vm_arch_init_platform_device(vm_t *v)
...
+    return 0;
+}
+
+int vm_enable_net(vm_t *v){
```
RinHizakura: Is this an arch-dependent function? If not, the function should be in `vm.c` like `vm_load_diskimg` which calls `virtio_blk_init_pci()`.

src/virtio-net.c
jimmylu890303: Sorry for my late. 
According to my understanding, this might be wrong. When calling `virtio_pci_get_notify_addr`, we obtain the address for the driver to notify the device. However, I think the structure layout is not similar to `virtio_pci_notify_data`. (Because I don't negotiate VIRTIO_F_NOTIFICATION_DATA, so only 2 byte for vqn)
```c
uint64_t virtio_pci_get_notify_addr(struct virtio_pci_dev *dev,
                                    struct virtq *vq)
{
    uint64_t base = PCI_HDR_READ(dev->pci_dev.hdr,
                                 PCI_BAR_OFFSET(dev->notify_cap->cap.bar), 32);
    uint64_t offset =
        dev->notify_cap->cap.offset +
        dev->notify_cap->notify_off_multiplier * vq->info.notify_off;
    return base + offset;
}
```
My understanding is that through `notify_off_multiplier` and `notify_off`, we can map two notification areas for RX and TX. We then simply poll these two areas for IO events.

image

src/arch/x86/vm.c
jimmylu890303: I already remove it

src/arch/x86/vm.c
jimmylu890303: This is not an arch-dependent function. So I move the `vm_enable_net` to /src/vm.c.
`vm_enable_net` will initialize the tap device and then register the virtio-net device to bus.
```c
int vm_enable_net(vm_t *v){
    if(!virtio_net_init(&v->virtio_net_dev))
        return -1;
    virtio_net_init_pci(&v->virtio_net_dev,&v->pci,
                        &v->io_bus,&v->mmio_bus);
    return 0;
}
```

src/virtio-net.c
RinHizakura: Yes. You don't negotiate `VIRTIO_F_NOTIFICATION_DATA` so we only need just 2 bytes for notify data, but what I care about is the "abstraction".

Let me ask in another way: If the host notifies tx vq, where of the memory space in kvm-host will update synchronously? Isn't it `vm.virtio_blk_dev.virtio_pci_dev.config.notify_data.next`?

jimmylu890303: The host (I think you mean the virtio device because the host emulates the backend hardware). So the question becomes: how does the device notify the virtqueue to do the transfer?

For example, when virtio-net finds that the tap device has data that needs to be transferred, it needs to get an available buffer from rx_virtq. The device reads the data, writes it to the virtqueue buffer, marks the buffer as used, and then uses a notify callback to send an IRQ to the guest.

Similarly, when the virtio-net device finds that the guest has data to transfer (by polling the notify_cap of the virtio device), the guest writes the data to an available virtqueue buffer from tx_virtq and marks it as available. The device can then get the available virtqueue, read the data from this buffer, and write it to the tap device to complete the transfer from guest to host.

RinHizakura: Hmm, that's not really what I'm asking. Thank you for the explanation of these high-level concepts, but I am just confused about how the code runs to emulate these things.

I mean, as the spec you post here, it says "The driver is sometime required to send an available buffer notification to the device... this notification involves sending the virtqueue number to the device.". Since we are emulating the device,  where can the emulator find the virtq number when the driver sends the notification?

RinHizakura: Based on your thinking, consider we have another new virtio device that needs 3 vqs, so you will define `notify_off_multiplier` to 2, and `vq0.notify_off=0,  vq1.notify_off=1, vq2.notify_off=2` to register 3 eventfds at 3 distinct addresses. According to my understanding, it will cause a problem in this case because we only preserve 4 bytes for notification.

So what I'm thinking here is whether to increase the entries of `struct virtio_pci_notify_data` under `struct virtio_pci_config`, and assign each vq to a different `virtio_pci_notify_data`. If my above thinking is correct.

jimmylu890303: > Hmm, that's not really what I'm asking. Thank you for the explanation of these high-level concepts, but I am just confused about how the code runs to emulate these things.
> 
> I mean, as the spec you post here, it says "The driver is sometime required to send an available buffer notification to the device... this notification involves sending the virtqueue number to the device.". Since we are emulating the device, where can the emulator find the virtq number when the driver sends the notification?

Sorry for my misunderstanding,
Actually in my implement, I don't directily use the vqn of notification data.
I just poll the notification address for io event, but emulator can using `dev->config.notify_data.vqn` to find the vqn.

jimmylu890303: > Based on your thinking, consider we have another new virtio device that needs 3 vqs, so you will define `notify_off_multiplier` to 2, and `vq0.notify_off=0, vq1.notify_off=1, vq2.notify_off=2` to register 3 eventfds at 3 distinct addresses. According to my understanding, it will cause a problem in this case because we only preserve 4 bytes for notification.
> 
> So what I'm thinking here is whether to increase the entries of `struct virtio_pci_notify_data` under `struct virtio_pci_config`, and assign each vq to a different `virtio_pci_notify_data`. If my above thinking is correct.

Yes, it might be a potential problem. However, when I traced notify_data.vqn, I found that the value is always 0, even when using the ping command (sending a packet should use tx_virtq and the guest should write the data to avail_virtq, so I would expect notify.vqn to be 1).

For increasing the entries in the struct `virtio_pci_notify_data`, I think it would require a redesign of the workflow, including changes to the virtio-pci logic. I need to think how to implement this improvement.

jimmylu890303: I think the KVM host uses `ioeventfd` for the notification mechanism, so we don't need to use the vqn.

When a Guest performs an IO operation that triggers a vm-exit, KVM first checks if an eventfd is registered for that address. If it is, KVM notifies through the eventfd and then allows the Guest to continue execution. This mechanism is useful for IO requests that are primarily for notifications rather than actual data transfer, such as the virtio-pci available buffer notification, thereby reducing the time overhead.

[ref1](https://patchwork.kernel.org/project/kvm/patch/1251028605-31977-23-git-send-email-avi@redhat.com/)
[ref2](https://patchwork.kernel.org/project/kvm/patch/1251028605-31977-23-git-send-email-avi@redhat.com/)

RinHizakura: Thank you for patiently explaining to me.  I got the point. With ioeventfd, you never use `virtio_pci_notify_data`, so the notification never truly writes the PCI MMIO space.

I'm just thinking if we are going to use ioeventfd for every notification, maybe we should consider removing `virtio_pci_notify_data`, but it is fine to keep this right now.

RinHizakura (approved):
I'm good with the changes now. Thanks for all the kindly reply.
 
Let' see if @jserv has any extra comment.

jimmylu890303:
> I'm good with the changes now. Thanks for all the kindly reply.
> 
> Let' see if @jserv has any extra comment.

Thanks for your review !

jimmylu890303:
@jserv , Please let me know your thoughts when you have a moment.

jserv (changes requested):
Rebase the latest `master` branch because the commit "Bump linux version to 6.1.91" was accidentally checked in as a non-functional change. See #35

jserv (changes requested):
Indent all files using `clang-format`. That is, run `find src -name '*.[ch]' | xargs clang-format-12 -i`. Do replace `clang-format-12` with [clang-format](https://clang.llvm.org/docs/ClangFormat.html) version 12 or later.

jimmylu890303:
I have pushed the latest version with the following updates:
* I rebased the latest master branch to remove the non-functional commit "Bump linux version to 6.1.91" as discussed in issue #35.
* I have reformatted all relevant files using clang-format. Specifically, I ran the command `find src -name '*.[ch]' | xargs clang-format-12 -i` with clang-format version 12.

jserv:
> 2. I have reformatted all relevant files using clang-format. Specifically, I ran the command find src -name '*.[ch]' | xargs clang-format-12 -i with clang-format version 12.

Warning: No newline at end of file `src/virtio-net.c`
See https://medium.com/@alexey.inkin/how-to-force-newline-at-end-of-files-and-why-you-should-do-it-fdf76d1d090e

Makefile
```diff
@@ -29,7 +29,7 @@ OBJS := \
 	pci.o \
 	virtio-pci.o \
 	virtq.o \
-	virtio-blk.o \
+	virtio-blk.o virtio-net.o\
```
jserv: Split into a new line.

src/virtio-net.c
```diff
@@ -0,0 +1,281 @@
...
+static struct virtq_ops virtio_net_ops[VIRTIO_NET_VIRTQ_NUM] = {
+    [VIRTQ_RX] =
+        {
+            .enable_vq = virtio_net_enable_vq_rx,
+            .complete_request = virtio_net_complete_request_rx,
+            .notify_used = virtio_net_notify_used_rx,
+        },
+    [VIRTQ_TX] = {.enable_vq = virtio_net_enable_vq_tx,
+                  .complete_request = virtio_net_complete_request_tx,
+                  .notify_used = virtio_net_notify_used_tx}};
```
jserv: Replace `.notify_used = virtio_net_notify_used_tx}};` with `.notify_used = virtio_net_notify_used_tx},};` and re-run `clang-format`.

src/virtio-net.c
```diff
@@ -0,0 +1,281 @@
...
+    if (virtio_net_dev->tapfd < 0) {
+        return false;
+    }
+    struct ifreq ifreq = {.ifr_flags = IFF_TAP | IFF_NO_PI};
+    strncpy(ifreq.ifr_name, TAP_INTERFACE, sizeof(ifreq.ifr_name));
+    if (ioctl(virtio_net_dev->tapfd, TUNSETIFF, &ifreq) < 0) {
+        fprintf(stderr, "failed to allocate TAP device: %s\n", strerror(errno));
+        return false;
+    }
+    assert(fcntl(virtio_net_dev->tapfd, F_SETFL,
```
jserv: Don't use `assert` because `NDEBUG` might be set.
> If the macro `NDEBUG` is defined at the moment `<assert.h>` was last included, the macro assert() generates no code, and hence does nothing at all.  It is not recommended to define NDEBUG if using `assert()` to detect error conditions since the software may behave non-deterministically.

jserv (changes requested):
Use `git rebase -i` to rework the git commits.
The following commits can be squashed and refined:
* Define VIRTIO_NET_IRQ 14
* Define VIRTIO_PCI_DEVICE_ID_NET macro
* Define macro in virtio-net.h
* Define struct virtio_net_dev and functions in virtio-net.h

Git commits are snapshots around logical units of change. Over time, commits should tell a story of the history of your repository and how it came to be the way that it currently is.
See https://github.blog/2022-06-30-write-better-commits-build-better-projects/

You don't have to mention the file names in the subject of git commit messages. Instead, summarize the functional changes.

jimmylu890303:
I have pushed the new version for review. Thank you, @jserv

src/virtio-net.c
```diff
@@ -0,0 +1,284 @@
+#include <assert.h>
```
jserv: Remove `<assert.h>` since it is not used.

src/main.c
```diff
@@ -93,6 +93,8 @@ int main(int argc, char *argv[])
         return throw_err("Failed to load initrd");
     if (diskimg_file && vm_load_diskimg(&vm, diskimg_file) < 0)
         return throw_err("Failed to load disk image");
+    if (vm_enable_net(&vm) < 0)
```
jserv, line 96: If virtio-net is not enabled, it still makes sense to initialize and launch this virtual machine via KVM. Thus, instead of returning, we can proceed.

jimmylu890303: I changed throw_err to fprintf. If `vm_enable_net` failed to enable virtio-net device, it will now handle the situation by logging an error message.
```c
if (vm_enable_net(&vm) < 0)
        fprintf(stderr, "Failed to Enable Virtio-Net Device\n");
```

src/virtio-net.c
jimmylu890303: Already removed it.

jserv (changes requested):
Unify the commit "Implement vm_enable_net API" and "Compile virtio-net.c into virtio-net.o during build process" into a single one since they are tended to the same purpose.

src/main.c
jserv: Refine the message as following:
> Failed to enable virtio-net device.

jimmylu890303: Ok ! Already done

jserv:
Thank @jimmylu890303 for contributing!

Issue #35, opened 2024-07-10
Fail to run kvm-host with Linux in version 6.1.91


RinHizakura:
In 0f31f1b, the kernel bumped to 6.1.91. I am not able to see any message in this version.
```
$ make check
Once the message 'Kernel panic' appears, press Ctrl-C to exit

```
Everything goes fine if I use the 6.1.35 kernel, so I believe this is not an issue that fully relates to my host PC.

I'm wondering if only I face the issue. Note that if we already built the kernel with `make check` before, we have to manually remove the `build/bzImage` to force `make check`  using the latest version of the kernel. Or we can clone a clean repo and build again. This could be another thing we have to fix.

RinHizakura:
Some information of my environment.

```
$ uname -r
6.5.0-35-generic

$ lscpu
Architecture:            x86_64
  CPU op-mode(s):        32-bit, 64-bit
  Address sizes:         46 bits physical, 48 bits virtual
  Byte Order:            Little Endian
CPU(s):                  32
  On-line CPU(s) list:   0-31
Vendor ID:               GenuineIntel
  Model name:            13th Gen Intel(R) Core(TM) i9-13900K
    CPU family:          6
    Model:               183
    Thread(s) per core:  2
    Core(s) per socket:  24
    Socket(s):           1
    Stepping:            1
    CPU max MHz:         5800.0000
    CPU min MHz:         800.0000
    BogoMIPS:            5990.40
    Flags:               fpu vme de pse tsc msr pae mce cx8 apic sep mtrr pge mca cmov pat pse36 clflush dts acpi mmx fxsr sse sse2 ss ht
                         tm pbe syscall nx pdpe1gb rdtscp lm constant_tsc art arch_perfmon pebs bts rep_good nopl xtopology nonstop_tsc cp
                         uid aperfmperf tsc_known_freq pni pclmulqdq dtes64 monitor ds_cpl vmx smx est tm2 ssse3 sdbg fma cx16 xtpr pdcm s
                         se4_1 sse4_2 x2apic movbe popcnt tsc_deadline_timer aes xsave avx f16c rdrand lahf_lm abm 3dnowprefetch cpuid_fau
                         lt epb ssbd ibrs ibpb stibp ibrs_enhanced tpr_shadow flexpriority ept vpid ept_ad fsgsbase tsc_adjust bmi1 avx2 s
[...]

PR #37, opened 2024-07-20
Fix regression when bumping Linux to 6.1.88

RinHizakura:
Starting from Linux v6.1.88 commit [f31f521](https://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git/commit/?h=v6.1.88&id=f31f521ad27e3e476a58c07bf0825c4441317dba), the first bootsec is complete removed and filled with 0xff to reserve for PE header. Since we load the full 512 bytes, those 0xff will break kvm-host. Instead, we only have to take the part of setup header.

### DIFF
--- a/src/arch/x86/vm.c
+++ b/src/arch/x86/vm.c
@@ -6,6 +6,7 @@
 #include <asm/e820.h>
 #include <linux/kvm.h>
 #include <linux/kvm_para.h>
+#include <stddef.h>
 #include <string.h>
 #include <sys/ioctl.h>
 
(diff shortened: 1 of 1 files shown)


RinHizakura:
Solve #35

src/arch/x86/vm.c
```diff
@@ -131,7 +132,9 @@ int vm_arch_load_image(vm_t *v, void *data, size_t datasz)
     void *kernel = ((uint8_t *) v->mem) + 0x100000;
 
     memset(boot, 0, sizeof(struct boot_params));
-    memmove(boot, data, sizeof(struct boot_params));
+    memmove((void *) boot + offsetof(struct boot_params, hdr),
```
jserv: When using void pointers in calculations, the behavior is undefined.
Consider the following:
```c
    memmove((void *) ((uintptr_t) boot + offsetof(struct boot_params, hdr)),
            (void *) ((uintptr_t) data + offsetof(struct boot_params, hdr)),
            sizeof(struct setup_header));
```

jserv: Add some comments to explain the rationale.

jserv:
Thank @RinHizakura for contributing!

PR #40, opened 2025-06-30
arm64: Enable virtio-net and add network scripts

otischung:
Enable POSIX timers to support ping’s POSIX timer syscalls:
- CONFIG_POSIX_TIMERS=y in configs/linux-arm64.config

Enable network stack and virtio-net in kernel config:
- CONFIG_NET, CONFIG_INET, CONFIG_IP_MULTICAST, CONFIG_IP_PNP
- CONFIG_IP_PNP_DHCP, CONFIG_NETDEVICES, CONFIG_VIRTIO
- CONFIG_VIRTIO_NET, CONFIG_VIRTIO_RING

Add virtio_net_dev to FDT generation:
- src/arch/arm64/vm.c: include virtio_net_dev in interrupt-map

Define VIRTIO_NET_IRQ=2 (SPI 34) in src/arch/arm64/desc.h

Add scripts for host and guest network setup:
- scripts/set-host-bridge.sh to configure TAP bridge and default route
- scripts/set-guest-route.sh to configure guest static IP and route

Update README.md with new scripts and build instructions

These changes have been validated on both a Raspberry Pi 5 Model B
(Revision 1.0) running Ubuntu 25.10 with Linux kernel 6.14.0 and an
NVIDIA Jetson Orin Nano Developer Kit running JetPack 6.2.1
(L4T 36.4.4) with Linux kernel 5.15.148-tegra.


visitorckw (changes requested):
Since patch 3 fixes a typo introduced in patch 1 of this PR, please squash patch 3 into patch 1 using git rebase -i to keep the history clean.

otischung:
Many thanks to @visitorckw.

I rebased the branch, interchanged patches 2 and 3, flagged patch 2 for squashing, and force-pushed the result. Please examine the changes at your convenience.

README.md
```diff
@@ -42,8 +44,63 @@ containing concatenated `bootsect.o + setup.o + misc.o + piggy.o`. `initrd` is t
 initial RAM disk image, which is an optional argument.
 `disk-image` is the path to disk image which can be mounted as a block device via virtio. For the reference Linux guest, ext4 filesystem is used for disk image.
 
+### Stop Emulator
```
jserv: It is meant to "stop" the emulation. Instead, the operation is dedicated to "exit" from the execution loop.

configs/linux-arm64.config
```diff
@@ -66,3 +66,15 @@ CONFIG_EXT4_FS=y
...
+CONFIG_IP_MULTICAST=y
+CONFIG_IP_PNP=y
+CONFIG_IP_PNP_DHCP=y
+CONFIG_NETDEVICES=y
+CONFIG_VIRTIO=y
+CONFIG_VIRTIO_NET=y
+CONFIG_VIRTIO_RING=y
+CONFIG_UNIX=y
+CONFIG_UNIX_SCM=y
+CONFIG_AF_UNIX_OOB=y
```
jserv: Confirm if they are needed or not.

otischung:
I’ve removed the unneeded ARM64 kernel configs (CONFIG_UNIX, CONFIG_UNIX_SCM, CONFIG_AF_UNIX_OOB) and renamed “Stop Emulator” to “Exit Emulator” in the README to better reflect its behavior. Please take a look and let me know if everything looks good for merge!

otischung:
### Update

The changes were validated on the NVIDIA Jetson Orin Nano Developer Kit running JetPack 6.2.1 (L4T 36.4.4) with Linux kernel version 5.15.148-tegra. Network functionality within the guest operating system was confirmed to operate correctly.

README.md
```diff
@@ -42,8 +44,63 @@ containing concatenated `bootsect.o + setup.o + misc.o + piggy.o`. `initrd` is t
...
+### Exit Emulator
+
 To exit kvm-host, press "Ctrl-A", release both keys, and then press "x".
 
+### Enable Static Route to Test the Guest VirtIO-Net Interface
+
+1. Start the kvm-host emulator. Once initialized, the TUN/TAP interface (for example, `tap0`) is visible in the output of `ip a`. The following is sample output from the host:
+
+   ```shell
+   ❯ ip a
```
jserv: Use `$` for the command line prompt.

README.md
```diff
@@ -42,8 +44,63 @@ containing concatenated `bootsect.o + setup.o + misc.o + piggy.o`. `initrd` is t
...
+          valid_lft xxxxxxsec preferred_lft xxxxxxsec
+       inet6 ...
+   11: tap0: <BROADCAST,MULTICAST> mtu 1500 qdisc noop state DOWN group default qlen 1000
+       link/ether 5a:1d:bd:2d:7c:1f brd ff:ff:ff:ff:ff:ff
+   ```
+
+2. Execute the shell script `./scripts/set-host-bridge.sh`, which configures a bridge by assigning the default route to 10.0.0.1, integrating the TUN/TAP interface into the bridge, and activating the network interfaces. The following is sample output of `ip a` from the host:
+
+   ```shell
+   ❯ ip a
```
jserv: Ditto. Use `$`.

otischung:
### Update

I’ve updated the README example to use the standard `$` prompt instead of `❯` for the `ip a` command.

Also, just a heads-up, the change requested by @visitorckw is **_no longer applicable_**, as the commit they referenced was _**removed**_ by a _**force‑push**_.

> Many thanks to @visitorckw.
> 
> I rebased the branch, interchanged patches 2 and 3, flagged patch 2 for squashing, and force-pushed the result. Please examine the changes at your convenience.

Let me know if you’d like any further tweaks. Thanks.

visitorckw (changes requested):
The From: header varies across patches (Otis Chung, Otis, otischung). Please ensure your git configuration (user.name) is consistent across all commits.

visitorckw (changes requested):
Later patches fix issues introduced in earlier patches of the same series. This "development history" should not be visible upstream. Please squash fixes into the original commits introducing the changes.

visitorckw (changes requested):
Patch 1 does too much (code, configs, scripts, docs). Please split it into logical units.

visitorckw (changes requested):
> Define VIRTIO_NET_IRQ=2 (SPI 34) in src/arch/arm64/desc.h

Please explain why IRQ 2 was chosen in the commit message. Is this mandated by the GIC or an arbitrary choice?

visitorckw (changes requested):
[PATCH 2/6] ("Fix the URL in README.md") looks like a standalone fix unrelated to virtio-net. Consider moving this to the beginning of the series or sending it as a separate trivial patch.

README.md
```diff
@@ -10,6 +10,11 @@ capable of running Linux kernel partially.
 
 ## Build and Run
 
+Ensure the following prerequisites are installed on Ubuntu 24.04:
```
visitorckw, line 13: I think docs should try to be distro-agnostic where possible, or label it as an example?

configs/linux-arm64.config
```diff
@@ -12,7 +12,7 @@ CONFIG_CC_OPTIMIZE_FOR_SIZE=y
 # CONFIG_MULTIUSER is not set
 # CONFIG_SYSFS_SYSCALL is not set
 # CONFIG_FHANDLE is not set
-# CONFIG_POSIX_TIMERS is not set
+CONFIG_POSIX_TIMERS=y
```
visitorckw, line 15: The commit message claims to "Enable ... virtio-net ... and the core networking stack". It does not mention POSIX Timers. Why is CONFIG_POSIX_TIMERS required for virtio-net?

scripts/set-host-bridge.sh
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+sudo ip link delete br0 || true
+sudo brctl addbr br0
+sudo ip addr add 10.0.0.1/24 dev br0
+sudo ip route add default via 10.0.0.1 dev br0
```
visitorckw, line 6: So this changes the host's default gateway to the bridge IP (10.0.0.1).
Wouldn't this make all internet traffic from the host dropped and kill the network connection of anyone who runs this script?

jserv:
Close in favor of #47

